PR #13, opened 2020-01-04
Allow choice of alternate compression (gzip, etc).

perlman:
While I realize this utility exists primarily as a stage of the slide-to-OME-TIFF workflow, it seems quite useful for converting strange formats into n5 for further processing.

This small patch simply allows for the selection of the default n5 compression techniques. My specific use-case is to support direct access from neuroglancer (raw/gzip).

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -87,6 +87,58 @@ public class Converter implements Callable<Void> {
   /** Scaling factor in X and Y between any two consecutive resolutions. */
   private static final int PYRAMID_SCALE = 2;
 
+  static class N5Compression {
+    enum CompressionTypes { blosc, bzip2, gzip, lz4, raw, xz };
+
+    private static Compression getCompressor(
+            CompressionTypes type,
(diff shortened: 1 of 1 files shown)


chris-allan:
Thanks very much for the contribution.

It would be preferred if we could use a class that represents the possible compression types. You can see an example of this style for the compression type configuration in `raw2ometiff`:

 * https://github.com/glencoesoftware/raw2ometiff/blob/master/src/main/java/com/glencoesoftware/pyramid/PyramidFromDirectoryWriter.java#L99
 * https://github.com/glencoesoftware/raw2ometiff/blob/master/src/main/java/com/glencoesoftware/pyramid/PyramidFromDirectoryWriter.java#L164

The strong advantage this has are (a) the potential options in `--help` are automatically generated and (b) all potential options can be interrogated programatically.

perlman:
How does this look? I went with an enum instead of extending List<String>. I also added a flag for compression level.  (I'm not remotely tied to any of the names -- and the style of code was an effort to comply with the rather strict gradle rules. :)

I started to go down the rabbit hole of a generic mapping, using the same SciJava search n5 uses (using reflections), but it got messy fast.

PR #15, opened 2020-02-14
Add support for reading Faas pyramids

melissalinkert:
This should enable conversion from Faas pyramids (pre-6.0.0 pyramid TIFFs) to 6.0.0+ pyramid OME-TIFFs.  ```PyramidTiffReader``` is copied from https://github.com/glencoesoftware/bioformats-private/blob/4655e5c/components/formats-gpl/src/loci/formats/in/PyramidTiffReader.java with minor style fixes required by Checkstyle.

Also adds a ```--exclude_readers``` option to remove any reader from the ```ClassList``` passed to ```ImageReader```.  By default, both ```MiraxReader``` and ```PyramidTiffReader``` are included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ Maximum tile dimensions are can be configured with the `--tile_width` and `--til
 resolution is no greater than 256x256.
 
 By default, two additional readers (MiraxReader and PyramidTiffReader) are added to the beginning of Bio-Formats' list of reader classes.
-These and any other readers can be removed from the list with the `--exclude_readers` option:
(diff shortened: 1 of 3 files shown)


chris-allan:
@melissalinkert: Have a look at chris-allan/bioformats2raw@dc986a6 for an alternative command line argument structure and let me know what you think.

melissalinkert:
Merged the ```--extra-readers``` change here.  Only issue I could see is that there wasn't a way to specify no extra readers, so 5d2348b is one way to fix that.

PR #16, opened 2020-03-08
Fix issues with little-endian source data

melissalinkert:
This should be tested together with https://github.com/ome/bioformats/pull/3519.  819e5ae makes it so that this should pick up the local build that includes https://github.com/ome/bioformats/pull/3519:

```
$ cd bioformats
$ git checkout melissalinkert/tiff-offset-overflow
$ mvn clean install
$ cd ../bioformats2raw
$ gradle clean build
```

/cc @douglasrennick, @chris-allan

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ repositories {
 }
 
 dependencies {
-    compile ('ome:formats-gpl:6.2.1')
+    compile ('ome:formats-gpl:6.4.0')
     compile ('info.picocli:picocli:3.9.6')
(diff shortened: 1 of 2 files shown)


melissalinkert:
Updated to use Bio-Formats 6.4.0 since that includes https://github.com/ome/bioformats/pull/3519.  The extra step of building and installing a local snapshot of Bio-Formats is no longer necessary.

chris-allan:
Verification performed by @douglasrennick.

PR #17, opened 2020-03-19
WIP: n5-aws-s3

perlman:
I've been playing around with using object stores for n5 volumes using [n5-aws-s3]( https://github.com/saalfeldlab/n5-aws-s3/). This PR is an initial (functional) proof-of-concept.

I'd like to gauge if it's useful to keep this in bioformats2raw, along with suggestions on how to refactor it. My thought is to use a separate class for I/O and/or a factory for creating the correct n5 and file objects.

I used [MockS3](https://github.com/adobe/S3Mock) for local testing.

[This work is with @mellertd at @TheJacksonLaboratory.]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,8 +26,10 @@ repositories {
 dependencies {
     compile ('ome:formats-gpl:6.4.0')
     compile ('info.picocli:picocli:3.9.6')
-    compile ('org.janelia.saalfeldlab:n5:2.1.2')
-    compile ('org.janelia.saalfeldlab:n5-blosc:1.0.0')
+    compile ('org.janelia.saalfeldlab:n5:2.1.6')
+    compile ('org.janelia.saalfeldlab:n5-blosc:1.0.1')
+    compile ('org.janelia.saalfeldlab:n5-aws-s3:3.0.0')
+    compile ('com.amazonaws:aws-java-sdk-s3:1.11.719')
 }
 
 applicationDistribution.from("$projectDir") {
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -250,14 +250,15 @@ public class Converter implements Callable<Void> {
 
   @Option(
           names = "--pyramid-name",
-          description = "Name of pyramid n5 (default: ${DEFAULT-VALUE})"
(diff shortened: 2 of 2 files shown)


melissalinkert:
Thanks for this @perlman, and apologies for not providing feedback sooner.  I can definitely see the use case here, but fundamentally I think would prefer to keep the S3-specific logic more separated, and ideally out of this repo.  One way to meet both goals might be to:

1. Define an interface that looks something like:

```
interface BlockStore {
    N5Reader createReader(String);
    N5Writer createWriter(String);
}
```

2. Add a ```BlockStore``` implementation (```FSStore```?) that constructs an ```N5FSReader```/```N5FSWriter```
3. Add options to this repo to specify the ```BlockStore``` implementation (```FSStore``` by default) for reading and writing
4. Add an implementation of ```BlockStore``` that constructs an ```N5AmazonS3Reader```/```N5AmazonS3Writer```  (```S3Store```?)

The end result being that something like this would read from disk and write an N5 structure to S3 without any S3-specific code in the ```Converter``` class:

```
bioformats2raw --write-store S3Store /path/to/file bucket-name.s3.us-west-1.amazon.com/n5-pyramid
```

Maybe 1,3,4 would be best placed in a separate repo that this (and https://github.com/glencoesoftware/raw2ometiff) can depend upon?  That's obviously a rough sketch of an idea, so happy to continue the conversation if you or anyone else have other thoughts.

perlman:
>  I can definitely see the use case here, but fundamentally I think would prefer to keep the S3-specific logic more separated, and ideally out of this repo. 

I agree completely, and am on board with the approach mentioned above.

What is your suggested course of action? Would it be useful for us (me) to refactor this into a PR, or to wait for your side to create a new repo for this? (I'm in no rush, this hack works for now.)

On a related note -- will bioformats2raw & raw2ometif be distributed through maven? I can certainly see myself extending them to make small behavioral changes.

joshmoore:
@perlman : I was going to give this a try based on https://github.com/bcdev/jzarr/pull/16#issuecomment-800376205 (or have you already got something in place?)

perlman:
> @perlman : I was going to give this a try based on [bcdev/jzarr#16 (comment)](https://github.com/bcdev/jzarr/pull/16#issuecomment-800376205) (or have you already got something in place?)

That would be great!

This proof-of-concept has clearly withered.

joshmoore:
@perman: see https://github.com/glencoesoftware/bioformats2raw/pull/89

PR #18, opened 2020-03-19
Bump n5 version to work with Java >= 11

perlman:
I spent longer than I care to admit debugging an error that was relating to my JAVA_HOME no longer pointing to JDK 1.8.

Newer Java releases work with n5 >= 2.1.4: https://github.com/saalfeldlab/n5/pull/59


melissalinkert:
Thanks, @perlman, this all makes sense.  Converted several files with and without this change, and the only difference in output is the version number written in the root attributes JSON.  Merging, and will open a similar pull request against raw2ometiff shortly.

PR #19, opened 2020-03-19
Flags to rename and pyramid.n5 and scaler paths

perlman:
neuroglancer's [n5 implementation ](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/n5/frontend.ts#L352) uses "s%d"  for each scale.

By using "s%d" instead (as an option) allows this an option.  The default remains "%d".

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -299,6 +299,12 @@ public class Converter implements Callable<Void> {
   public void convert()
       throws FormatException, IOException, InterruptedException
   {
+    if (!pyramidName.equals("pyramid.n5") || !scaleFormatString.equals("%d")) {
+      LOGGER.info("Output will be incompatible with raw2ometiff " +
+              "(pyramidName: {}, scaleFormatString: {})",
+              pyramidName, scaleFormatString);
+    }
+
     Cache<TilePointer, byte[]> tileCache = CacheBuilder.newBuilder()
         .maximumSize(maxCachedTiles)
         .build();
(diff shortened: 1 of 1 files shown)


melissalinkert:
Thank you for contributing this, @perlman.  The changes look fine to me, and increasing compatibility with other tools makes sense.

The only thing I can think of that might be nice is to somehow warn that using using non-default values for these options will break compatibility with https://github.com/glencoesoftware/raw2ometiff (the second half of the workflow this repo was designed for).  That could be something in the option descriptions, or a message logged at ```WARN``` if non-default values are detected.

perlman:
Your request makes sense -- there is no harm in being verbose here. I'll update with something later this week.

perlman:
I've updated to add the bare minimum warning to the help text ("[Can break compatibility with raw2ometiff]") and a logger.INFO message when the values are changed.

I looked into using a picocli ArgGroup to cluster the flags which might break raw2ometiff support, and possibly use that as hook to see if the options are changed, but that feature was introduced in picocli 4.0 (which is a breaking update from 3.9.6). 🤷‍♂️

melissalinkert:
Latest commits look fine, and all works for me in a final test.  Merging to get these new commands included.

I don't have a problem with upgrading picocli in a follow-up pull request if needed.  We've upgraded from 3.9.3 to 4.1.4+ elsewhere with no issues.

joshmoore:
Hi @perlman. Not sure of exactly which repo to discuss on, but do you have any suggestions how we might synchronize rollout of the naming changes from https://github.com/zarr-developers/zarr-specs/issues/50 for neuroglancer, bioformats2raw, etc.?

perlman:
Hi @joshmoore!

I've been following that thread and almost every minor concern I may have had has been addressed in the comments. I'll go through again and see if I have anything to add.

I'm excited by the idea of a synchronized rollout of the new format (and am fully aware that it  obliterates my motivation or *this* specific PR.) I think an issue on each repo, referencing the proposal, would be good place to start?

Issue #25, opened 2020-04-17
Scope expansion beyond whole slide imaging


chris-allan:
The initial intent of this repository and the complementary `raw2ometiff` tool was primarily the conversion of proprietary whole slide imaging file formats, such as MRXS, into OME-TIFF. Over the past few months there have been various external (#13, #19) contributions and internal discussions surrounding expanding this scope. Currently the data layout after conversion consists of the following files:

* `METADATA.ome.xml` -- A dump of the OME-XML generated by Bio-Formats produced when initializing the Bio-Formats reader on the source file
* `pyramid.n5` -- A 3 dimension (X, Y, C) N5 pyramid with levels described by ascending stringified N5 datasets as loosely described on zarr-developers/zarr-specs#50
* `LABELIMAGE.jpg` -- A JPEG compressed version of the source file Bio-Formats series which corresponds to the "Label image" (optional, only Bio-Formats series 1)
* `<series_no>.jpg` -- A JPEG compressed version of additional Bio-Formats series beyond 1

If coming from `isyntax2raw` the conversion may also consist of the following additional files:
* `METADATA.json` -- A JSON encoded dump of the source file metadata
* `MACROIMAGE.jpg` -- A JPEG compressed version of the source file Bio-Formats series which corresponds to the "Macro image"

This conversion data layout is obviously quite singular in its intent. Firstly, we would like to expand to cover:

1. Multi-series input beyond whole slide imaging in a generic fashion
2. Dimensions (Z, T) beyond the three currently being considered

This would mean:

1. Using a generic `data.n5` or similar name for the N5 data
1. Dispensing with the secondary JPEG compressed `LABELIMAGE.jpg`, `<series_no>.jpg`, etc. files and encoding *all* series from the source file in ascending stringified order
1. Expanding the number of dimensions to 5 (X, Y, C, Z, T) following the Bio-Formats declared, and OME-XML recorded, dimension order setting their size to 1 if missing entirely
1. Adding version metadata in a defined location to aid downstream consumers
1. Adding an option to optionally force the dimension order

Consequently, the layout should reflect the following:

```
data/
├── 0    # Series 0
|   ├── 0    # Full-sized array
|   ├── 1    # Scaled down 0, e.g. 0.5; for images, in the X&Y dimensions
|   ├── 2    # Scaled down 1, ...
|   ├── 3    # Scaled down 2, ...
|   └── 4    # Etc.
├── 1    # Series 1
|   └── 0    # Etc.
[...]

PR #28, opened 2020-04-20
Multi series and ZCT support

chris-allan:
Implementation of multi-series and ZCT support in accordance with #25

- [x] Using a generic data.n5 or similar name for the N5 data
- [x] Dispensing with the secondary JPEG compressed LABELIMAGE.jpg, <series_no>.jpg, etc. files and encoding all series from the source file in ascending stringified order
- [x] Expanding the number of dimensions to 5 (X, Y, C, Z, T) following the Bio-Formats declared, and OME-XML recorded, dimension order setting their size to 1 if missing entirely
- [x] Adding version metadata in a defined location to aid downstream consumers
- [x] Adding an option to optionally force the dimension order

* Using `data.n5` lacking other dissenting opinions
* Option to force dimension order is `--dimension-order`
* Layout version is currently set to `1` and is available on the root (`/`) group at the `bioformats2raw.layout` key

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,6 @@
 plugins {
     id 'application'
     id 'eclipse'
-    id 'java'
     id 'checkstyle'
 }
 
@@ -14,7 +13,6 @@ targetCompatibility = 1.8
 
 repositories {
     jcenter()
-    mavenCentral()
     maven {
         url 'https://artifacts.openmicroscopy.org/artifactory/repo/'
     }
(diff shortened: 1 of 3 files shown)


chris-allan:
@joshmoore: Ready for a quick review pass. Probably needs to have bytewise tests added to ensure the right data is going to the right place now.

chris-allan:
Directory issues should be resolved now. I just have a couple more tests I want to make to ensure that the tile encoding is correct.

chris-allan:
Handling cases along the right most and bottom most edge where the dimensions are not evenly divisible by the tile size is currently broken. Opened saalfeldlab/n5-zarr#2 upstream as a fix.

/cc @joshmoore, @melissalinkert

chris-allan:
Dimension orders now preserved with command line option for overriding.

chris-allan:
Layout version now available in the metadata. Largely now feature complete vs. the specification on #25 and ready for final review before merging.

chris-allan:
Our version of n5-zarr (`0.0.3-SNAPSHOT`) now going to repo.glencoesoftware.com, depending on it here and tests are now all enabled.

melissalinkert:
All looks much better.  It might make sense to save the value of ```dimensionOrder``` (if not null) to the ```METADATA.ome.xml``` file for consistency, but I realize that might need further discussion so not a blocker to merging.

Issue #11, opened 2019-12-05
'?m' not a supported value of 'class ome.xml.model.enums.UnitsLength'


olatarkowska:
I have converted MRXS to n5 using https://github.com/glencoesoftware/bioformats2raw, now trying to convert to ome-tiff but I am facing issue with the following:

```/opt/raw2ometiff-0.1.0$ bin/raw2ometiff /nfs/MRXS/HB_BRC2083.n5 /nfs/MRXS/HB_BRC2083.ome.tif
2019-12-05 11:47:37,277 [main] INFO  c.g.p.PyramidFromDirectoryWriter - Creating tiled pyramid file /nfs/MRXS/HB_BRC2083.ome.tif
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.janelia.saalfeldlab.n5.CompressionAdapter (file:/opt/raw2ometiff-0.1.0/lib/n5-2.1.3.jar) to field java.lang.reflect.Field.modifiers
WARNING: Please consider reporting this to the maintainers of org.janelia.saalfeldlab.n5.CompressionAdapter
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
2019-12-05 11:47:37,472 [main] INFO  org.perf4j.TimingLogger - start[1575546457002] time[469] tag[initialize]
Exception in thread "main" picocli.CommandLine$ExecutionException: Error while calling command (com.glencoesoftware.pyramid.PyramidFromDirectoryWriter@47dbb1e2): java.lang.RuntimeException: loci.formats.FormatException: Could not parse OME-XML
	at picocli.CommandLine.execute(CommandLine.java:1180)
	at picocli.CommandLine.access$800(CommandLine.java:141)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:1367)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:1335)
	at picocli.CommandLine$AbstractParseResultHandler.handleParseResult(CommandLine.java:1243)
	at picocli.CommandLine.parseWithHandlers(CommandLine.java:1526)
	at picocli.CommandLine.call(CommandLine.java:1786)
	at picocli.CommandLine.call(CommandLine.java:1710)
	at com.glencoesoftware.pyramid.PyramidFromDirectoryWriter.main(PyramidFromDirectoryWriter.java:260)
[...]

PR #29, opened 2020-04-22
Docker: DEBIAN_FRONTEND=noninteractive

manics:
When I built the Docker image yesterday it hung with a prompt requesting the timezone.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ ARG BUILD_IMAGE=gradle:6.2.1-jdk8
 #
 FROM ${BUILD_IMAGE} as build
 USER root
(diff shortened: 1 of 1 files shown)


chris-allan:
I'm certainly by no means a Docker authority but it seems to me we might want to deal with the standard problems of timezone assignment, locale, etc. that appear to occur every time with minimal install Ubuntu Docker images rather than hacking this. See also, #11.

joshmoore:
The alternative would be to take the timezone from the building environment?

joshmoore:
Unless someone has a suggestion, I'd be for getting this workaround in since I need to keep this local to prevent docker build from hanging.

PR #30, opened 2020-04-22
Add additional scale format string arguments

chris-allan:
Adds the capability to manipulate the scale format string through an additional argument CSV in line with the requirements detailed in #25. It is now possible to do all sorts of crazy things with the N5/Zarr hierarchy.

The format string follows the Java formatter style:

 * https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html
 * https://mkyong.com/java/java-string-format-examples/

There is probably an extensive amount of documentation that is possible here the implementation feels sound.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ repositories {
 
 dependencies {
     implementation 'ome:formats-gpl:6.4.0'
-    implementation 'info.picocli:picocli:3.9.6'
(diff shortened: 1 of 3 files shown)


chris-allan:
Cool. Going to merge this then and if we find we need additional stuff for our use case or @joshmoore needs additional changes for IDR conversions we can get those in as part of a separate PR.

PR #33, opened 2020-04-29
Fix edge effects during downsampling

chris-allan:
On read, files in Zarr format must be de-padded. Source from IDR:

 * http://idr.openmicroscopy.org/webclient/?show=image-179706
 * http://idr.openmicroscopy.org/webclient/img_detail/179706/

Before:

test

After:

test2

/cc @will-moore

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -591,13 +591,18 @@ public class Converter implements Callable<Void> {
           pathName, datasetAttributes, gridPosition
         ).toByteBuffer();
 
-        int length = blockWidth * bytesPerPixel;
+        int destLength = blockWidth * bytesPerPixel;
+        int srcLength = destLength;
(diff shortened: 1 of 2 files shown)


src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -591,13 +591,18 @@ public void write(int series)
           pathName, datasetAttributes, gridPosition
         ).toByteBuffer();
 
-        int length = blockWidth * bytesPerPixel;
+        int destLength = blockWidth * bytesPerPixel;
+        int srcLength = destLength;
+        if (fileType == FileType.zarr) {
+          // n5/n5-zarr does not de-pad on read
+          srcLength = activeTileWidth;
```
melissalinkert, line 598: Not sure I understand - why doesn't this need to be multiplied by ```bytesPerPixel```?

chris-allan: Oversight; good catch.

Issue #37, opened 2020-05-17
Aperio SVS file conversion shows 3 errors


dsudar:
First of all thanks for this great set of tools.

I'm trying to use them to convert Aperio .svs files into ome.tif (so we can distribute them in a standard format and so I can stick additional metadata into the OME-XML header). These are regular H&E SVS images in RGB just like the public test image (https://downloads.openmicroscopy.org/images/SVS/77917.svs) with a label and a macro image. However there are a number of problems with that:
1) the original RGB image is down-converted to a single-channel image.
2) the label image comes through in the resulting ome.tif but the macro image isn't there. showinf also complains about: "Error untangling IFDs; the OME-TIFF file may be malformed (IFD #2 missing)"
3) a lot of the useful metadata from the svs file is not carried forward to the resulting ome.tif.

Thanks,
Damir

melissalinkert:
Thanks for reporting this, @dsudar.  Could you please let us know which versions of bioformats2raw and raw2ometiff you were using?

dsudar:
I was using bioformats2raw-0.2.0-SNAPSHOT and raw2ometiff-0.1.1. Sorry, I hadn't seen that there's a 0.2.2 version of both. I'll upgrade and try again.

dsudar:
Hi @melissalinkert Again apologies for reporting on an old release. I tried the latest release 0.2.1 of both programs. Issues 1 and 2 are solved. Per issue #3, it would still be useful to carry all the svs original metadata forward to the resulting ome.tif if possible.

BUT, a new issue sneaked in: at a number of the subresolution levels the image data is corrupted. You can see that with 77917.svs image when you convert to ome.tif, load the result into OMERO and visualize with PathViewer. I'm enclosing a screenshot.
![Capture](https://user-images.githubusercontent.com/10214074/82234659-8d5ffe80-98e6-11ea-8c6a-94843b2285f9.JPG)

melissalinkert:
I can reproduce both problems now.  #38 should resolve the missing metadata issue, and https://github.com/glencoesoftware/raw2ometiff/pull/27 should fix the corrupt image data.

dsudar:
Great! I'll try to build a copy with those PRs and check it out. Thanks!!

dsudar:
Quick test on both PRs. All works great for me. Thanks.

chris-allan:
Fixes released downstream in 0.2.3:

[...]

PR #38, opened 2020-05-18
Preserve original metadata key/value pairs in OME-XML

melissalinkert:
See #37

```showinf -nopix -noflat``` on the output of bioformats2raw + raw2ometiff should show the same metadata table as ```showinf -nopix -noflat``` on the original file.


chris-allan:
Would it be possible to add a fake file test case for this as we have done for the Zarr layouts?

Issue #49, opened 2020-07-07
query for version


dsudar:
Please add a --version flag to allow querying what version of this program is currently installed. Also, nice if that can query and report on the version of the Bio-Formats library incorporated into the current version.

chris-allan:
#50 has been merged and released in `0.2.5`.

PR #50, opened 2020-07-08
Add --version option

melissalinkert:
The Gradle build will generate a ```version.properties``` file that contains the bioformats2raw and Bio-Formats version numbers.  Running ```bioformats2raw --version``` will print both values and exit.

See #49

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,8 +25,9 @@ repositories {
     }
 }
 
+def bfVersion = '6.4.0'
 dependencies {
-    implementation 'ome:formats-gpl:6.4.0'
+    implementation "ome:formats-gpl:${bfVersion}"
     implementation 'info.picocli:picocli:4.2.0'
     implementation 'com.univocity:univocity-parsers:2.8.4'
     implementation 'org.janelia.saalfeldlab:n5:2.2.0'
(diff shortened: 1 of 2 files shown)


chris-allan:
My strong preference here would be to use the JAR implementation version as we have with the microservices and stitcher:

 * https://github.com/glencoesoftware/omero-ms-image-region/blob/master/build.gradle#L93
 * https://github.com/glencoesoftware/omero-ms-image-region/blob/master/src/main/java/com/glencoesoftware/omero/ms/image/region/ImageRegionMicroserviceVerticle.java#L266
 * https://github.com/glencoesoftware/sticher/blob/master/src/main/java/com/glencoesoftware/stitcher/Main.java#L34

Is there also a reason why we wouldn't use the `FormatTools` constants in the stitcher way as well?

build.gradle
```diff
@@ -26,7 +26,7 @@ repositories {
 }
 
 dependencies {
-    implementation 'ome:formats-gpl:6.4.0'
+    implementation "ome:formats-gpl:6.4.0"
```
chris-allan, line 29: Was there a reason for this?

melissalinkert: Not any more; changed back in 7106854.

melissalinkert:
7106854 writes the version numbers to stdout instead of using slf4j, which should look better.  If you prefer to see those logged with slf4j, then we either need to change the logging configuration overall (not ideal), or override the default configuration when ```--version``` is used.

chris-allan:
@melissalinkert: I'm fine with either. STDOUT works for me if you prefer.

melissalinkert:
stdout is completely fine with me.

PR #53, opened 2020-08-04
Add --pixel-type option to change pixel type during conversion

melissalinkert:
Currently only supports float and double as input types, and int8/uint8/int16/uint16 as output types.

The specified pixel type should be an OME-XML ```PixelType```, e.g. ```--pixel-type uint16```.


melissalinkert:
Is this still a feature that we want?  If so, I'll resolve the latest conflicts, but if not let's just close it.

PR #56, opened 2020-08-20
Warn if the output path already exists

melissalinkert:
See https://github.com/ome/omero-insight/issues/155


chris-allan:
Had occurred to me that this warning will probably be lost in the current amount of output. Shall we make it fatal and add `--overwrite` or similar which will remove the output directory entirely if it exists before initiating conversion?

chris-allan:
Fixes for the misconfigured test case (my fault) are in ac78fd8 and a fix for the recursive deletion are in f7faeb3 on `master`.

PR #60, opened 2020-09-24
Add uint32 support

melissalinkert:
Only simple downsampling is support for uint32 data, as OpenCV does not have a corresponding data type.  Tested with open source ```gatan/qa-11392``` sample file.


chris-allan (changes requested):
Let's add a test case for int32 and uint32.

PR #61, opened 2020-09-29
Catch and log errors when loading OpenCV native libraries

melissalinkert:
See https://github.com/openpnp/opencv/issues/44.  c6741f0 catches ```Throwable``` so that the default simple downsampling should continue to work even if OpenCV cannot be used.

5dce02c changes how OpenCV is loaded as suggested in https://github.com/openpnp/opencv/issues/44.  If using an OpenCV downsampling type (e.g. ```--downsample-type AREA```) with Windows + Java 14 still doesn't work, this commit could be reverted.

/cc @erindiel

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -390,8 +390,7 @@ public class Converter implements Callable<Void> {
       return null;
     }
 
-    nu.pattern.OpenCV.loadShared();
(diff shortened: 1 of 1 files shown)


erindiel (commented):
Confirming that conversion runs successfully on Windows + Java 14 with and without OpenCV downsampling type `--downsample-type AREA`

Images converted to OME TIFF with raw2ometiff and quality checked. Downsampling differed between OpenCV downsampling types as expected.

PR #64, opened 2020-10-13
Temporarily disable open source artifactory

chris-allan:
Power outage at UoD has resulted in a significant outage. We have our own mirrors, let's use them.


chris-allan:
/cc @kkoz 

You can use this style to fix our microservice builds.

PR #65, opened 2020-10-19
First try at configurable background/fill color

melissalinkert:
Applies to .mrxs data only for now.  See commit message for https://github.com/glencoesoftware/bioformats2raw/commit/ad5b8b55e2a7a0752512607123d66e1c782ab931

/cc @erindiel

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -350,10 +350,10 @@ public class Converter implements Callable<Void> {
 
   @Option(
           names = "--fill-value",
-          description = "Default value to fill in for missing tiles" +
+          description = "Default value to fill in for missing tiles (0-255)" +
                         " (currently .mrxs only)"
   )
-  private volatile Byte fillValue = null;
+  private volatile Short fillValue = null;
 
   @Option(
           arity = "0..1",
@@ -405,6 +405,10 @@ public class Converter implements Callable<Void> {
       System.out.println("Version = " + version);
       System.out.println("Bio-Formats version = " + FormatTools.VERSION);
       return null;
+    }
+
+    if (fillValue != null && (fillValue < 0 || fillValue > 255)) {
+      throw new IllegalArgumentException("Invalid fill value: " + fillValue);
     }
 
     loadOpenCV();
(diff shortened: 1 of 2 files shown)


src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -487,6 +494,9 @@ public void convert()
       Memoizer memoizer;
       try {
         reader = (IFormatReader) readerClass.getConstructor().newInstance();
+        if (reader instanceof MiraxReader) {
+          ((MiraxReader) reader).setFillValue(fillValue);
```
joshmoore, line 498: I know it says so in the doc string but maybe fail fast if fillValue is set and this isn't a MiraxReader?

melissalinkert: See b1efcd2.  For right now we just care about .mrxs data, but long term the general idea would be:

- push ```setFillValue``` and ```getFillValue``` (or whatever we agree the API should be) up to ```IFormatReader``` and ```FormatReader```
- replace ```Arrays.fill(..., (byte) 0)``` with ```Arrays.fill(..., getFillValue())``` in appropriate readers
- revert b1efcd2 and call ```setFillValue``` on all readers

chris-allan:
Needs to have conflicts resolved and a definitive :+1: from @erindiel before merging.

erindiel:
Conversions of both Fluo and RGB MRXS files are correctly filled with black/white background without having to specify `--fill-value`.

Setting `--fill-value` does produce the appropriately custom-filled-background output, but I note that the expected input seems to be signed, rather than unsigned. For example, I can set 
`--fill-value=-100` and `fill-value=100`
but
`-fill-value=255` produces error `Invalid value for option '--fill-value': '255' is not a byte`
I would expect the acceptable range to be 0-255 (but this also introduces the question of what happens with 16-bit images?)

melissalinkert:
Conflicts resolved in 9e5ccd6.  978381c switches the expected range of ```fillValue``` to unsigned byte, so ```--fill-value=255``` should work.

Any fill value that is specified is applied to every background byte (not every background pixel), so on a 16-bit image setting ```--fill-value=255``` results in pixel values of 65535.  If we need to be able to specify a pixel value instead, that's certainly possible, it's just a bit more work.

erindiel (approved):

chris-allan:
Apologies. After the merge of #68, will need conflicts resolved again and then we can merge.

PR #69, opened 2020-11-13
Initial OME Zarr HCS implementation

melissalinkert:
See https://github.com/ome/omero-ms-zarr/pull/75

Automatically uses HCS layout if a plate is present in the reader's ```MetadataStore```, unless the ```--no-hcs``` option is used.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -991,7 +991,7 @@ public class Converter implements Callable<Void> {
               Map<String, Object> image = new HashMap<String, Object>();
               int plateAcq = field.getPlateAcquisitionIndex();
               image.put("acquisition", String.valueOf(plateAcq));
-              image.put("path", plateAcq);
+              image.put("path", String.valueOf(field.getFieldIndex()));
               imageList.add(image);
             }
           }
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -275,6 +275,12 @@ public class Converter implements Callable<Void> {
   )
   private volatile List<String> readerOptions = new ArrayList<String>();
 
+  @Option(
+          names = "--no-hcs",
+          description = "Turn off HCS writing"
+  )
+  private volatile boolean noHCS = false;
+
   /** Scaling implementation that will be used during downsampling. */
   private volatile IImageScaler scaler = new SimpleImageScaler();
 
@@ -306,6 +312,8 @@ public class Converter implements Callable<Void> {
 
   /** Current number of tiles processed at the current resolution. */
   private volatile AtomicInteger nTile;
+
+  private List<HCSIndex> hcsIndexes = new ArrayList<HCSIndex>();
 
   @Override
   public Void call() throws Exception {
@@ -434,8 +442,17 @@ public class Converter implements Callable<Void> {
         seriesCount = v.getSeriesCount();
         IMetadata meta = (IMetadata) v.getMetadataStore();
 
+        if (!noHCS) {
+          noHCS = meta.getPlateCount() == 0;
+        }
+
         for (int s=0; s<meta.getImageCount(); s++) {
           meta.setPixelsBigEndian(true, s);
+
+          if (!noHCS) {
+            HCSIndex index = new HCSIndex(meta, s);
+            hcsIndexes.add(index);
+          }
         }
 
         String xml = getService().getOMEXML(meta);
@@ -453,6 +470,10 @@ public class Converter implements Callable<Void> {
       }
       finally {
         readers.put(v);
+      }
+
+      if (!noHCS) {
+        scaleFormatString = "%d/%d/%d/%d/%d/%d";
       }
 
       for (int i=0; i<seriesCount; i++) {
@@ -515,12 +536,23 @@ public class Converter implements Callable<Void> {
       Integer series, Integer resolution)
   {
     List<Object> args = new ArrayList<Object>();
-    args.add(series);
-    args.add(resolution);
-    if (additionalScaleFormatStringArgs != null) {
-      String[] row = additionalScaleFormatStringArgs.get(series);
-      for (String arg : row) {
-        args.add(arg);
+    if (!noHCS) {
+      HCSIndex index = hcsIndexes.get(series);
+      args.add(index.getPlateIndex());
+      args.add(index.getPlateAcquisitionIndex());
+      args.add(index.getWellRowIndex());
+      args.add(index.getWellColumnIndex());
+      args.add(index.getFieldIndex());
+      args.add(resolution);
+    }
+    else {
+      args.add(series);
+      args.add(resolution);
+      if (additionalScaleFormatStringArgs != null) {
+        String[] row = additionalScaleFormatStringArgs.get(series);
+        for (String arg : row) {
+          args.add(arg);
+        }
       }
     }
     return args.toArray();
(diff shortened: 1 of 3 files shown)


src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -886,6 +923,80 @@ public void saveResolutions(int series)
...
+    }
+
+    // assumes only one plate defined
+    int plate = 0;
+    Map<String, Object> plateMap = new HashMap<String, Object>();
+
+    plateMap.put("name", meta.getPlateName(plate));
+
+    List<Map<String, Object>> columns = new ArrayList<Map<String, Object>>();
+    for (int c=0; c<meta.getPlateColumns(plate).getValue(); c++) {
```
sbesson, line 938: After 30min of conversion of a plate, the call ended with a `java.lang.NullPointerException` while trying to write the HCS metadata. Now I understand better why https://github.com/ome/bioformats/pull/3638 is required :)

Two  questions:
- does this mean progress is blocked on having https://github.com/ome/bioformats/pull/3638 available in an mainline release?
- or do we need to implement some fallback anyways for the readers where this metadata  might be lacking?

sbesson:
Tested the last commit with the IDR Bio-Formats JAR and `--file_type zarr` against https://idr.openmicroscopy.org/webclient/?show=plate-2551. Conversion successfully completed in ~45min and generated a structure very close to the samples published and announced in https://blog.openmicroscopy.org/file-formats/community/2020/12/01/zarr-hcs/.

A few comments:

- the first two levels under the top-level `data.zarr` are now the rows and columns and contain 8 elements and 12 elements respectively as expected
- this representation uses 0-based indexing for all the groups (vs letter-based indexing and 1-based indexing for the samples generated by OME-Zarr)
- as per the above, a next step will be to test these generated datasets against some of the clients we have used for driving the HCS spec - this might lead us to reviewing some assumptions
- are the more usual A-Z and 1-12 names for wells/stored stored anywhere in the OME metadata or is it an OMERO concept? 
- each well contains 12 images but only 6 are referenced in the `.zattrs` (as expected)

  ```
  (base) [sbesson@pilot-zarr1-dev 2551.zarr]$ less data.zarr/0/0/
  0/       1/       10/      11/      2/       3/       4/       5/       6/       7/       8/       9/       .zattrs  .zgroup
  (base) [sbesson@pilot-zarr1-dev 2551.zarr]$ cat data.zarr/0/0/.zattrs 
  {"well":{"images":[{"path":0,"acquisition":"0"},{"path":0,"acquisition":"0"},{"path":1,"acquisition":"1"},{"path":1,"acquisition":"1"},{"path":2,"acquisition":"2"},{"path":2,"acquisition":"2"},{"path":3,"acquisition":"3"},{"path":3,"acquisition":"3"},{"path":4,"acquisition":"4"},{"path":4,"acquisition":"4"},{"path":5,"acquisition":"5"},{"path":5,"acquisition":"5"}]}}(base) [sbesson@pilot-zarr1-dev 2551.zarr]$ 
  ```
-  more generally, as I am trying to diff JSON files with different formatting conventions, any immediate suggestion of the most appropriate tool?

melissalinkert:
> are the more usual A-Z and 1-12 names for wells/stored stored anywhere in the OME metadata or is it an OMERO concept?

The metadata contains 0-based indexes and optional ```RowNamingConvention``` and ```ColumnNamingConvention``` on ```Plate```: https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#NamingConvention.  The typical well names would have to be calculated accordingly.

> more generally, as I am trying to diff JSON files with different formatting conventions, any immediate suggestion of the most appropriate tool?

Not really, unfortunately.  I run the bioformats2raw JSON through a pretty-printer (https://github.com/makamaka/JSON-PP) as needed, but that doesn't completely solve the problem.

melissalinkert:
Which IDR Bio-Formats commit specifically were you using?  I am having trouble finding a version that detects 6 fields and plate acquisitions without throwing an exception.

sbesson:
Sorry, I should have precised, I build `bioformats2raw` from 998e4cf7b6e53d550a04560239bfa8f18a0828af with the following local modification

```
(base) [sbesson@pilot-zarr1-dev bioformats2raw]$ git diff
diff --git a/build.gradle b/build.gradle
index 54006ec..8af170e 100644
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ repositories {
 }
 
 dependencies {
-    implementation 'ome:formats-gpl:6.5.1'
+    implementation 'idr:formats-gpl:0.6.5'
     implementation 'info.picocli:picocli:4.2.0'
     implementation 'com.univocity:univocity-parsers:2.8.4'
     implementation 'org.janelia.saalfeldlab:n5:2.2.0'
```

melissalinkert:
Ah, OK.  That's the same version I was using, but a different dataset (I had https://downloads.openmicroscopy.org/images/Flex/idr0001/).

As far as I can tell by looking at the original data for ```idr0001-graml-sysgro/20151116-verified/JL_120731_S6A```, there should in fact be 12 images per well - 2 fields x 6 plate acquisitions.  ```showinf -nopix -omexml``` with https://github.com/idr/bioformats/tree/IDR-0.9.0 shows 2 ```WellSamples``` for each well in each plate acquisition, and the .flex file names match up with those dimensions.  The JSON in https://github.com/glencoesoftware/bioformats2raw/pull/69#issuecomment-740646559 lists 12 images per well, but the paths are wrong for half of them.  This is fixed in 48dccc8.

sbesson:
After double checking the data, you are absolutely right the dataset contains multiple fields of view per well for each acquisition 

```
$ ls -alh /nfs/bioimage/drop//idr0001-graml-sysgro/20151116-verified/JL_120731_S6A/Meas_01\(2012-07-31_10-41-12\)/
total 17G
drwxrwx---. 2 6166 idrnfs  20K May 15  2016 .
drwxrwx---. 8 6166 idrnfs 4.0K May 15  2016 ..
-rw-rw----. 1 6166 idrnfs  88M May  4  2016 001001001.flex
-rw-rw----. 1 6166 idrnfs  88M May  4  2016 001001002.flex
-rw-rw----. 1 6166 idrnfs  88M May  4  2016 001002001.flex
-rw-rw----. 1 6166 idrnfs  88M May  4  2016 001002002.flex
```

Unfortunately this means the dataset is not directly comparable to the data exported from IDR as only the first field of view of each well seems to have been imported. Proposing to discuss at the next Formats meeting the next steps for validating this PR as well as the handling of empty rows/columns.

chris-allan:
@sbesson: We would like to get this in for the forthcoming 0.3.0 release. Is there anything further you'd like to discuss/test here beforehand?

melissalinkert:
Conflicts resolved.

sbesson (approved):
@chris-allan thanks for the heads up. 

Summarizing briefly, most of my testing has been focused on the use case of plates with multiple acquisitions. This is primarily because this was the scenario we were trying to work in terms of specification. Unfortunately for the reasons discussed above, it is not straightforward to compare  the output of bioformats2raw vs omero-cli-zarr. 

A potential test might to try and convert the other sample plates discussed in the public [OME-Zarr HCS spec blog post](https://blog.openmicroscopy.org/file-formats/community/2020/12/01/zarr-hcs/) using `bioformats2raw` and compare the output with the public Zarr representations on S3. Given the preliminary testing as well as my current capacity I certainly don't feel a compelling reason to block the upcoming minor release though.

Instead I'd propose to schedule this conversion test using `bioformats2raw 0.3.0` and capture issues separately.

chris-allan:
@melissalinkert: This PR is ready to have conflicts resolved prior to merge.

melissalinkert:
Conflicts resolved.

Issue #70, opened 2020-11-18
Add -series flag (a la bfconvert) to select from a multi-image input file


dsudar:
Many WSI images contain a slide label image and a macro image that are currently converted along with the main image. Sometimes those little images contain PHI data or other proprietary data that should be stripped out during conversion. Possible to add that new feature similar to bfconvert's -series flag?

melissalinkert:
#71 adds ```-s```/```--series```, which should work exactly like ```bfconvert -series```.

dsudar:
Works great. Thanks Melissa!

chris-allan:
Closing as the feature from is now shipping as of 0.3.0rc1.

PR #71, opened 2020-11-19
Add --series option to convert a single series

melissalinkert:
See #70.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -128,9 +128,11 @@ public class Converter implements Callable<Void> {
 
   @Option(
     names = {"-s", "--series"},
-    description = "Series index, if a single series should be converted"
+    arity = "0..1",
+    split = ",",
+    description = "Comma-separated list of series indexes to convert"
   )
-  private volatile int singleSeries = -1;
+  private volatile List<Integer> seriesList = new ArrayList<Integer>();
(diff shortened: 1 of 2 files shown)


sbesson (approved):
I have tested this PR in a slightly different context than the one described in #70, in my case a multi-resolution IMS file with downsampling alongside the Z dimension which is currently split into separate images rather than separate resolutions.


Without any options, the N5 folder contains 5 top-level folders and the `METADATA.ome.xml` lists four `Image` elements are in the top-level. With `--series 0`, the N5 directory contains one top-level folder and only one `Image` is listed in `METADATA.ome.xml`.

Happy to see this in a public release of `bioformats2raw`, test the `raw2ometiff` aspect of the pipeline and provide some feedback.

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -124,6 +126,12 @@
   )
   private volatile Integer pyramidResolutions;
 
+  @Option(
+    names = {"-s", "--series"},
+    description = "Series index, if a single series should be converted"
+  )
+  private volatile int singleSeries = -1;
```
chris-allan, line 133: Might be an idea to just go with `series` here of type `Integer` to allow for nullability and ease of detection if set.

chris-allan (changes requested):
How do we feel about allowing more than one series to be selected via delimeted or providing the same argument multiple times?

Issue #74, opened 2020-12-03
Command-line option for changing the log level


andreasg123:
Currently, one can only choose among the log levels `debug` or `info`. As `info` is still very verbose, it would be great to be able to select `warn` or other log levels.

melissalinkert:
Thanks, @andreasg123.  One way of doing this is proposed in #75.

andreasg123:
Thanks for addressing this so quickly. It would be great if you could add the same to [raw2ometiff](https://github.com/glencoesoftware/raw2ometiff).

joshmoore:
Included in https://github.com/glencoesoftware/raw2ometiff/pull/48

PR #75, opened 2020-12-08
Add --log-level option to set the specific logging level

melissalinkert:
Default is ```INFO```, and ```--debug``` sets to ```DEBUG```.  ```--log-level``` takes an optional
argument which is one of the logback levels:

http://logback.qos.ch/apidocs/ch/qos/logback/classic/Level.html

If the argument is not specified or is an invalid level name, ```DEBUG``` is used.

See #74.

This shouldn't change the behavior of ```--debug```, and means that ```--log-level``` and ```--debug``` can't both be specified.  There are probably other ways to do the same thing, this is just one option.  Whatever we agree on here should happen in raw2ometiff as well.


chris-allan:
Seems fine to me. We can go ahead and use the same style for raw2ometiff.

PR #76, opened 2020-12-09
Add a few notes on when to use --max_workers

melissalinkert:
Based upon recent comments from @manics and @sbesson.


manics (approved):
Thanks!

PR #80, opened 2021-02-22
Nested chunk storage

joshmoore:
Allow using N5-style chunk separators ("/") to prevent an obscene number (10s of millions) of chunks in a single directory. This was found while working on https://github.com/ome/ngff-latency-benchmark/pull/14 since directory metadata methods like file listings became the bottleneck.

This depends on https://github.com/bcdev/jzarr/pull/17 for the underlying logic.

To access these from zarr-python, a release of https://github.com/zarr-developers/zarr-python/pull/699 will be necessary.


joshmoore:
Sounds good.

joshmoore:
Updated, but we'll need a decision on the jzarr versioning before this will go green.

PR #81, opened 2021-02-22
Retain existing resolutions and add chunking in Z

dgault:
This work was intended initially for the NGFF benchmarking file generation and is built on top of the open jzarr PR.

Resolutions:
- automatically writes existing resolutions if they are present in the file rather than downsampling
- may make sense to add a flag to trigger the behaviour

3d chunking:
- adds the option to pass a parameter to enable chunking in Z
- tests need to be added
- if the original file is downsampled in Z then the additional resolutions will appear as separate images rather than resolutions. This is due to a Bio-Formats limitation

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -109,7 +109,7 @@
         <!-- Checks for Size Violations.                    -->
         <!-- See https://checkstyle.org/config_sizes.html -->
         <!--<module name="MethodLength"/>-->
-        <module name="ParameterNumber"/>
+        <!--<module name="ParameterNumber"/>-->
 
         <!-- Checks for whitespace                               -->
         <!-- See https://checkstyle.org/config_whitespace.html -->
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -109,7 +109,7 @@
         <!-- Checks for Size Violations.                    -->
         <!-- See https://checkstyle.org/config_sizes.html -->
         <!--<module name="MethodLength"/>-->
-        <!--<module name="ParameterNumber"/>-->
+        <module name="ParameterNumber"/>
 
         <!-- Checks for whitespace                               -->
         <!-- See https://checkstyle.org/config_whitespace.html -->
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -327,7 +327,7 @@ public class Converter implements Callable<Void> {
 
   )
   private volatile boolean noRootGroup = false;
-  
+
   @Option(
       names = "--use-existing-resolutions",
       description = "Use existing sub resolutions from original input format"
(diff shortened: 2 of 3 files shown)


chris-allan:
#78 has had `master` merged in and been updated. In the next day or so we aim to merge that so this PR can then have `master` be merged into it or be rebased and force pushed.

chris-allan:
#78 now in, ready for conflict resolution in this PR.

joshmoore:
@dgault : handled the progress bar conflicts in https://github.com/joshmoore/bioformats2raw/pull/new/retain-resol but haven't handled the ByteBuffer compilation issue yet

```
> Task :compileJava FAILED
/opt/bioformats2raw/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java:1037: error: incompatible types: byte[] cannot be converted to ByteBuffer
    writeBytes(zarr, shape, offset, chunkAsBytes.toByteArray());
                                                            ^
Note: Some input files use or override a deprecated API.
```

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -1099,33 +1214,38 @@ public void saveResolutions(int series)
...
-                    resolution, plane, xx, yy, width, height, t);
-                }
-                finally {
-                  if (pb != null) {
-                    pb.step();
+            for (int l=0; l<scaledDepth; l+=chunkDepth) {
+              final int zz = l;
+              int depth = (int) Math.min(chunkDepth, scaledDepth - zz);
+              for (int i=0; i<ctIndices[l].size(); i++) {
+                final int plane = ctIndices[l].get(i);
```
melissalinkert, line 1221: Is there a reason to have ```ctIndices``` instead of just looping over C and T (anticipating chunking in those dimensions as well)?

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -975,6 +993,64 @@ private void processTile(
     writeBytes(zarr, shape, offset, tileBuffer);
   }
 
+  private void processChunk(
+      int series, int resolution, int plane, int xx, int yy, int zz,
+      int width, int height, int depth)
```
melissalinkert, line 998: Does it make sense to take ```int[] offset, int[] shape``` instead of the six position/length parameters?  That would be easier to extend to C/T chunks, and would eliminate the need for the checkstyle configuration change.

config/checkstyle/checkstyle.xml
```diff
@@ -109,7 +109,7 @@
         <!-- Checks for Size Violations.                    -->
         <!-- See https://checkstyle.org/config_sizes.html -->
         <!--<module name="MethodLength"/>-->
-        <module name="ParameterNumber"/>
+        <!--<module name="ParameterNumber"/>-->
```
melissalinkert, line 112: I'd prefer to have this change reverted, see comment in ```processChunk```.

melissalinkert (changes requested):
Definitely needs an option to control whether existing resolutions are kept, as we certainly have data that contains not-quite-right pyramids.  My preference is to keep the original behavior of discarding existing resolutions by default, though happy to discuss.

Also definitely needs tests for chunking and keeping existing resolutions.  I think I have convinced myself that this works as intended, but would be good to formalize.

My assumption in some of the comments is that chunking in C and T is now only a matter of time.  If that's not accurate, then clarifying some of the Z-specific logic via comments instead of adding full C/T awareness may be fine.

If ```processTile``` is now unused, it can also be removed here for clarity.

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -1011,7 +1096,9 @@ public void saveResolutions(int series)
       LOGGER.info("Using {} pyramid resolutions", resolutions);
       sizeX = workingReader.getSizeX();
       sizeY = workingReader.getSizeY();
+      sizeZ = workingReader.getSizeZ();
       imageCount = workingReader.getImageCount();
+      imageCount /= sizeZ;
```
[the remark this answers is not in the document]
dgault: Image count is being used to calculate the total number of tiles on line https://github.com/glencoesoftware/bioformats2raw/pull/81/files#diff-156407914fc2609796db393053797aa5ce0446652d302b8bb7a6885b6adfa642R1176

As that calculation now takes into account chunked Z the imageCount needs reduced to not include Z. It might be clearer to create a new variable further on down with the count of C and T planes.

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -1069,7 +1172,19 @@ public void saveResolutions(int series)
       nTile = new AtomicInteger(0);
       tileCount = (int) Math.ceil((double) scaledWidth / tileWidth)
           * (int) Math.ceil((double) scaledHeight / tileHeight)
+          * (int) Math.ceil((double) scaledDepth / chunkDepth)
```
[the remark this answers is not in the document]
dgault: chunkDepth is correct in this case (same as the height and width). The Math.ceil will round up to 1 if it is less than scaledDepth

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -1224,8 +1232,20 @@ public void saveResolutions(int series)
...
-                    processChunk(series, resolution, plane, xx, yy, zz,
-                                 width, height, depth);
+                    int[] shape = {1, 1, 1, height, width};
+                    int[] offset;
+                    IFormatReader reader = readers.take();
+                    try {
+                      String o = new StringBuilder(
+                          dimensionOrder != null? dimensionOrder.toString()
+                          : reader.getDimensionOrder()).reverse().toString();
+                      shape[o.indexOf("Z")] = depth;
```
melissalinkert, line 1242: I'm not sure I understand why ```o.indexOf("Z")``` is used here when line 1025 above assumes ```shape[2]``` is the Z size.

dgault:
Not yet ready for review, changes for default behaviour still to come.

melissalinkert (approved):
I am fine with this being merged as-is.  Two minor points that I can address in a follow-up PR are:

- ```--chunk_depth``` doesn't break compatibility with raw2ometiff, but ```--use-existing-resolutions``` can (so command descriptions just need to be changed)
- an extra test for chunking with SizeC or SizeT greater than 1 and non-default dimension order would be helpful

joshmoore:
:+1:

melissalinkert:
Last two commits look fine, thanks @dgault.  OK to merge from my side.

chris-allan:
Only small piece of feedback is that we should probably align the `--chunk_depth` command line option with the symbol separator style (`_` vs `-`) used by the rest of the options. Will merge this as is and do that in a separate commit on `master`.

chris-allan:
Looking more closely, we've already got a series of mismatches. I'll leave it alone for now and we can come back to it for later breaking releases.

chris-allan:
Thanks again for this work, @dgault, it's really important functionality for us all to have!

PR #82, opened 2021-02-23
OpenCV library loading cleanup

chris-allan:
Couple small fixes to help when debugging OpenCV issues and avoid spurious OpenCV library load issues in logs.


chris-allan:
Note for the future,: upgrading to the OpenCV 4.3.0-3 release from the openpnp group is not currently viable due to glibc dependency issues on Linux. It would de facto drop CentOS 7 and Ubuntu 16.04 support from bioformats2raw which isn't really acceptable.

melissalinkert (approved):

PR #85, opened 2021-03-04
Retrieve OME-XML from ChannelSeparator, not base reader

melissalinkert:
This gives ChannelSeparator the chance to copy channel metadata in the case of RGB source data.

Without this change, ```METADATA.ome.xml``` would have a single ```Channel``` with ```SamplesPerPixel``` set to 3.  This doesn't match how the pixel data is actually converted.  raw2ometiff will then change the ```SamplesPerPixel``` to 1 and insert 2 more channels (https://github.com/glencoesoftware/raw2ometiff/blob/master/src/main/java/com/glencoesoftware/pyramid/PyramidFromDirectoryWriter.java#L442) with no metadata.

With this change, ```METADATA.ome.xml``` should have 3 ```Channel```s with ```SamplesPerPixel``` set to 1 and identical names.  The output of raw2ometiff should preserve this, so that there are no unnamed channels.

Only RGB input data (i.e. ```reader.isRGB()``` returns true) is affected.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -440,12 +440,13 @@ public class Converter implements Callable<Void> {
       memoizer.setFlattenedResolutions(false);
       memoizer.setMetadataFiltered(true);
       memoizer.setMetadataStore(createMetadata());
-      memoizer.setId(inputPath.toString());
-      memoizer.setResolution(0);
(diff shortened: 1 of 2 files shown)


chris-allan:
Let me know if you think chris-allan/bioformats2raw@c8bab4fe9bf4814a22bd293223f81ced8bfdf6f1 is enough @melissalinkert and I'll update the branch and merge.

melissalinkert:
That test looks fine to me, thanks @chris-allan.

Issue #79, opened 2021-01-20
0.3.0


chris-allan:
Follow on from discussion on #78. Major version checklist:

- [x] Bump `master` to 0.3.0-SNAPSHOT
- [x] Document the removal of N5 support and merge #78
- [x] Document changes to command line arguments
- [x] Bump layout version
- [x] Update layout based on suggestions from #78 (will have an effect on `isyntax2raw` and `raw2ometiff`)
- [x] Document existing layout and changes to layout

/cc @melissalinkert, @joshmoore

chris-allan:
`master` bumped to 0.3.0-SNAPSHOT in 030bfd285907ee95d471cf5313001c6ddb260467.

chris-allan:
0.3.0 has been released. Closing.

PR #86, opened 2021-03-10
Change default dimension order to TCZYX

melissalinkert:
See https://ngff.openmicroscopy.org/0.1/#image-layout, #79, https://github.com/glencoesoftware/bioformats2raw/pull/78#discussion_r591488638

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,17 @@ Usage Changes
 Versions 0.2.6 and prior supported both N5 and Zarr output using the `--file_type` option.
 This option is not present in 0.3.0 and later, as only Zarr output is supported.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -82,6 +82,17 @@ Usage Changes
 Versions 0.2.6 and prior supported both N5 and Zarr output using the `--file_type` option.
 This option is not present in 0.3.0 and later, as only Zarr output is supported.
 
+Versions 0.2.6 and prior used the input file's dimension order to determine the output
```
chris-allan, line 85: Might be an idea to clarify the difference in how we write down array index order since this will be the first instance that many people will be exposed to this setting. We're taking a "Bio-Formats" style dimension order on the command line, correct?

PR #90, opened 2021-03-17
Add no ome meta export option

MaximLippeveld:
Hi,

I've recently encountered some OOM errors caused by the construction of the OME-XML object. Since this is not strictly necessary for the functioning of the tool, I added an option that simply lets the user skip OME XML exporting.

Maxim

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -305,6 +305,12 @@ public class Converter implements Callable<Void> {
   )
   private volatile boolean noHCS = false;
 
+  @Option(
+          names = "--no-ome-meta-export",
+          description = "Turn off OME metadata exporting"
+  )
+  private volatile boolean noOMEMeta = false;
+
   /** Scaling implementation that will be used during downsampling. */
   private volatile IImageScaler scaler = new SimpleImageScaler();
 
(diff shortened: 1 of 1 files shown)


chris-allan:
Needs to have `master` merged in and conflicts resolved. We would also ask that a test case be added to verify the option. You can see examples doing this in `ZarrTest`.

chris-allan (changes requested):
Couple small things:

1. Since the use of `--no-ome-meta-export` __will__ break `raw2ometiff` and that tool is the primary consumer of intermediates generated by this project let's make sure that a statement in the usage help similar to that already in place for `--scale-format-string` and `--dimension-order`. I'd suggest "[Will break compatibility with raw2ometiff]".
1. You should be able to reduce memory even further by conditionally using `DummyMetadata` when `--no-ome-meta-export` is active instead of an `OMEXMLMetadata` in `createMetadata()`. This will prevent the creation of any DOM objects during metadata processing in the Bio-Formats `setId()` phase.

MaximLippeveld:
Added the statement to the help.

I tried to implement the usage of `DummyMetadata`, but ran into some problems. The `metadata` object is explicitly cast to `OMEXMLMetadata` in order to access the  `resolveReferences` method (for exmple on line 483 of `Converter`). If we just skip these calls when `noOMEMeta` is true, the `testNoOMEMetaExportOption` test fails due to the `outputPath` not being created properly.

Any pointers on how to tackle this?

melissalinkert:
Is the expectation that ```--no-ome-meta-export``` implies ```--no-hcs```?  If not, then using ```DummyMetadata``` is not viable in general.

Changing this: https://github.com/glencoesoftware/bioformats2raw/blob/3f6e63f26e6c873f0683148ec0aa12ea4d7796b1/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java#L463 to ```memoizer.setOriginalMetadataPopulated(!noOMEMeta)``` might be an easier/safer way to save some memory in that case.

chris-allan:
Thanks for your contribution as well as your patience and perseverance with this PR up until merging @MaximLippeveld!

MaximLippeveld:
> Thanks for your contribution as well as your patience and perseverance with this PR up until merging @MaximLippeveld!

My pleasure! Thanks for following up.

PR #91, opened 2021-03-17
Add "row_index" and "column_index" for each well

melissalinkert:
See https://github.com/ome/ngff/pull/24/commits/0c286908a4fbaceb2c622b21e5a52cc041eea7c0

Only acquired wells are included in the ```wells``` list, so this combined with the specification commit is a proposal for how to better handle sparse plates without requiring well path parsing.

/cc @kkoz


chris-allan:
Looks good. Let's add some test cases for some sequential, non-sequential and odd Well layouts?

PR #92, opened 2021-03-22
Remove "data.zarr" from output hierarchy by default.

melissalinkert:
This is the less-breaking option, which still keeps ```--pyramid-name``` so that the old behavior can be preserved with ```--pyramid-name data.zarr```.  Could also remove ```--pyramid-name``` entirely if anyone thinks it's no longer necessary.

The output directory can be anything right now, there is no requirement for it to end in ```.zarr```.  Should a ```.zarr``` ending be enforced?

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -221,7 +221,7 @@ public class Converter implements Callable<Void> {
           description = "Name of pyramid (default: ${DEFAULT-VALUE}) " +
                   "[Can break compatibility with raw2ometiff]"
   )
-  private volatile String pyramidName = "data.zarr";
+  private volatile String pyramidName = null;
 
   @Option(
           names = "--scale-format-string",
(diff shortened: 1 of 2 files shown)


joshmoore:
My personal use case is to be able to use the series name mapping to go from a multi-series PFF to:

```
${outputPath}/omero-image-id-123.zarr
${outputPath}/omero-image-id-456.zarr
```
where `outputPath` does *not* end in Zarr. Turning off the OME-XML generation, should let me write to S3 directly. I notice writing this that means I also don't want the top-level `.zgroup`...

To your points:
 * I don't particularly see the need for everything to end in Zarr if the user doesn't specify as much, though I could certainly live with a WARNing.
 * `--pyramid-name` might be more equivalent to `--wrapping-group`? (The presence of the OME-XML within the wrapping group might be considered `undefined`)

chris-allan:
All makes sense to me, @joshmoore. I think with this PR and #90 those requirements would be satisfied. Anything else you'd like to add before we merge?

joshmoore:
> https://github.com/glencoesoftware/bioformats2raw/pull/92#issuecomment-804974623 Anything else you'd like to add before we merge?

I haven't had a play and happy for it to go in as it stands, but what I would investigate next is if there's an (easy) way to omit the top-level `.zgroup` if `pyramid-name` isn't set.

chris-allan:
8ed20dc2394c128fc54370a172371d91702c1865 no longer assumes the presence of a root group for the purposes of conversion and allows it to be disabled if desired based on discussions with @joshmoore. The bioformats2raw --> raw2ometiff workflow requires the presence of such a group as it is the hosting point for the bioformats2raw layout metadata.

PR #98, opened 2021-04-13
If --no-hcs is passed, remove all HCS metadata from the OME-XML

sbesson:
See also a simlar change done in `ImageConverter` in https://github.com/ome/bioformats/pull/3259

In the scenario where the input data is HCS but the `--no-hcs` flag is passed, the `plate/row/col` groups are not created in the Zarr  layout. This change also removes the HCS metadata from the OME-XML file exported under the `OME` group.

In the scenario where both `--no-hcs` and `--series` are passed, this should prevent reference issues e.g. if the file is reconverted into OME-TIFF.

A simple example can be tested using a fake file e.g. 

```
bioformats2raw --no-hcs --series 0 "test&plates=1&plateRows=2&plateCols=2.fake" out
```

The HCS conversion unit tests have been expanded to test the existence of a `Plate` element with or without the `--no-hcs` flag.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -492,6 +492,9 @@ public class Converter implements Callable<Void> {
           for (int i=0; i<toRemove; i++) {
             root.removeImage(root.getImage(0));
           }
+          for (int i=0; i<meta.getPlateCount(); i++) {
+            root.removePlate(root.getPlate(0));
+          }
           meta.setRoot(root);
         }
         else {
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -492,8 +492,10 @@ public class Converter implements Callable<Void> {
           for (int i=0; i<toRemove; i++) {
             root.removeImage(root.getImage(0));
           }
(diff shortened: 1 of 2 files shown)


melissalinkert:
As written now, I'd expect this to break if ```--no-hcs``` is not specified with ```--series n```.  I think I'm OK with forcing ```--no-hcs``` for now (as long as it's documented), but I guess the question is if there is ever a reason to convert a subset of wells while keeping the whole plate relationship.

sbesson:
> I guess the question is if there is ever a reason to convert a subset of wells while keeping the whole plate relationship.

At least I don't know of a concrete use case but others might. A technical complication worth mentioning is that extracting only the relevant bit of the `Plate` metadata matching  the specified series is not completely trivial.

As hinted in the description, my workflow involves `bioformats2raw --no-hcs --series 0` and the issue is that this removes the HCS Zarr groups but keeps the HCS metadata which I consider as inconsistent. If we are not sure about the implications, would it make sense to update my changes and make them conditional to `--no-hcs` for now? And we can review whether `--series` implies `--no-hcs` separately.

sbesson:
The last commits update the logic to remove the `Plate` metadata only is `--no-hcs` is passed and add some coverage in the unit tests. The PR description has also been updated to reflect the new behavior.

melissalinkert:
This reads fine to me.  Might also make sense to add a ```--no-hcs``` test to raw2ometiff after this is merged.

chris-allan:
Test case added downstream in glencoesoftware/raw2ometiff#58.

PR #99, opened 2021-04-19
Fix nested/no-nested values

joshmoore:
When passing either `--nested` or `--no-nested` (rather than depending
on the default), the logic was reverseed. From https://picocli.info/#_negatable_options

> "When a negatable option is true by default, give it the negative name."


melissalinkert:
Makes sense to me.

PR #100, opened 2021-04-23
mrxs: simplify compressed tile buffer length calculation

melissalinkert:
/cc @erindiel @emilroz

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/MiraxReader.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/MiraxReader.java
@@ -109,8 +109,6 @@ public class MiraxReader extends FormatReader {
   private int yTiles;
   private int divPerSide;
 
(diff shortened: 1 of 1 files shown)


chris-allan:
@melissalinkert: Integer overflow?

melissalinkert:
Not integer overflow.  For larger slides, tiles for the largest resolution can be split across multiple files.  The first tile offset in each file is almost always the same, so ```firstLevelOffsets``` had duplicate entries.  The "next offset" calculation would only look for the first matching offset in ```firstLevelOffsets```, not necessarily the one corresponding to the correct file.  The tile lengths are all stored alongside the tile offsets in ```Index.dat```, so it's better to just use those than try to calculate the correct length.

chris-allan:
Validated on fluorescence and H&E data we have available in the test suite. Merging.

PR #105, opened 2021-06-10
Allow 3DHISTECH data to be initialized from Slidedat.ini file

melissalinkert:
This handles the case when the .mrxs file was lost or never saved.

/cc @erindiel


melissalinkert:
After further testing, I don't think this needs an extra type checking commit.  The two readers I was concerned about were ```FakeReader``` (which uses ```.fake.ini```) and ```HitachiReader``` (which uses INI files with .txt extension).  Neither have a problem with the change here, as only files named ```Slidedat.ini``` will be picked up by ```MiraxReader```.

erindiel (approved):
Confirmed that conversion worked as expected to both .zarr and .ome.tiff for datasets missing .mrxs files. Thank you @melissalinkert .

PR #114, opened 2021-08-20
Allow to override min_size via command-line argument

sbesson:
The current implementation stops downsampling the image as soon as both X and Y dimensions are smaller than 256. In some cases, we want this cut-off values to be configurable.

This commit makes the minimal change by keeping the default as 256 but allowing to override it via command-line option.

Tests are also added to cover the min-size arguments with various scenarios

### DIFF
--- a/src/test/java/com/glencoesoftware/bioformats2raw/test/ZarrTest.java
+++ b/src/test/java/com/glencoesoftware/bioformats2raw/test/ZarrTest.java
@@ -1206,6 +1206,29 @@ public class ZarrTest {
     }
   }
 
+  /**
+   * Convert an image with the --min-size option.
+   */
+  @Test
+  public void testMinSizeOption() throws Exception {
+    input = fake();
+    assertTool("--min-size", "32");
+
+    ZarrGroup z = ZarrGroup.open(output.resolve("0").toString());
+    List<Map<String, Object>> multiscales = (List<Map<String, Object>>)
+            z.getAttributes().get("multiscales");
+
+    Map<String, Object> multiscale = multiscales.get(0);
+    List<Map<String, Object>> datasets =
+              (List<Map<String, Object>>) multiscale.get("datasets");
+
+    String last_path = (String) datasets.get(datasets.size()  - 1).get("path");
+    ZarrArray last_array = z.openArray(last_path);
+
+    assert last_array.getShape()[0] < 32;
+    assert last_array.getShape()[1] < 32;
+  }
+
   private void checkPlateGroupLayout(Path root, int rowCount, int colCount,
     int fieldCount, int x, int y)
     throws IOException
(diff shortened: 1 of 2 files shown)


src/test/java/com/glencoesoftware/bioformats2raw/test/ZarrTest.java
```diff
@@ -1222,11 +1222,11 @@ public void testMinSizeOption() throws Exception {
...
 
-    String last_path = (String) datasets.get(datasets.size()  - 1).get("path");
-    ZarrArray last_array = z.openArray(last_path);
+    String lastPath = (String) datasets.get(datasets.size()  - 1).get("path");
+    ZarrArray lastArray = z.openArray(lastPath);
 
-    assert last_array.getShape()[0] < 32;
-    assert last_array.getShape()[1] < 32;
+    assert lastArray.getShape()[0] < 32;
+    assert lastArray.getShape()[1] < 32;
```
melissalinkert, line 1229: Should this be checking array elements 3 and 4 (Y and X)? I was expecting this test to fail, since the default fake image size is 512x512, which means the smallest resolution should be 32x32.

sbesson: :+1: updated in https://github.com/glencoesoftware/bioformats2raw/pull/114/commits/a8989140de62595eedacea5631c0eca46d87c2d9 to use `assertEqualArrays` on the return of `getShape`

melissalinkert (approved):
:+1: tests look much nicer now, thanks.

I do wonder though if the ```--min-size 30``` tests expose undesired behavior; an 8x16 image doesn't match with the ```Minimum size of the largest XY dimension``` being 30. That isn't an issue introduced in this PR, so I'm happy for this to be merged and we can discuss the correct behavior separately.

sbesson:
Heh. I was hesitant to start this conversation but since you brought it up, I am also not convinced by the `--min-size` terminology which is confusing as proven by your example. To be precise, the parameter is actually a maximum allowed size rather than a minimal size but it applies to the smallest resolution.

If we agreed on the thumbnail nomenclature, something like `--max-thumbnail-size` would probably work. Should we brainstorm for additional terms before we introduce the option in the command-line?

melissalinkert:
I'm fine with ```--max-thumbnail-size```.  ```--max-apex-size``` is the only other option that comes to mind (to avoid confusion between the smallest resolution and any separate thumbnail image), but that's probably not any clearer overall.

chris-allan:
My inclination is to call this `--target-min-size` or similar and refer to what the "target" is explicitly in the documentation. As mentioned this is a desired minimum size of longest side of the highest level of the pyramid rather than a hard limit.

chris-allan:
75b2863d274e01c5c41f7145b9674e6ad94d93d7 conceptually looks great to me. Will defer to @melissalinkert for anything else here.

melissalinkert:
Looks good to me too, thanks!

PR #118, opened 2021-10-15
Fix plate detection to ignore empty plates

melissalinkert:
HCS metadata will now only be written by default if there is at least one Plate, which contains at least one Well, which has at least one WellSample that links to an existing Image ID.  This means that empty plates will be ignored for the purposes of writing Zarr, but will be preserved in METADATA.ome.xml.

For datasets that have only empty/invalid plates, the default ```bioformats2raw``` behavior with this PR should match ```bioformats2raw --no-hcs``` without this PR.  No test (yet?) as creating an empty plate via fake files isn't possible.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -550,7 +550,7 @@ public class Converter implements Callable<Void> {
         ((OMEXMLMetadata) meta).resolveReferences();
 
         if (!noHCS) {
-          noHCS = meta.getPlateCount() == 0;
+          noHCS = !hasValidPlate(meta);
         }
         else {
           ((OMEXMLMetadata) meta).resolveReferences();
@@ -1742,6 +1742,31 @@ public class Converter implements Callable<Void> {
     finally {
       imageReader.close();
     }
+  }
+
+  /**
+   * Check if the given metadata object contains at least one plate
+   * with at least one well that links to an image.
+   *
+   * @param meta metadata object
+   * @return true if a valid plate exists, false otherwise
+   */
+  private boolean hasValidPlate(IMetadata meta) {
+    List<String> images = new ArrayList<String>();
+    for (int i=0; i<meta.getImageCount(); i++) {
+      images.add(meta.getImageID(i));
+    }
+    for (int p=0; p<meta.getPlateCount(); p++) {
+      for (int w=0; w<meta.getWellCount(p); w++) {
+        for (int ws=0; ws<meta.getWellSampleCount(p, w); ws++) {
+          if (images.contains(meta.getWellSampleImageRef(p, w, ws))) {
+            return true;
+          }
+        }
+      }
+      LOGGER.warn("Encountered invalid plate #{}", p);
+    }
+    return false;
   }
 
   private static Slf4JStopWatch stopWatch() {
--- a/src/test/java/com/glencoesoftware/bioformats2raw/test/ZarrTest.java
+++ b/src/test/java/com/glencoesoftware/bioformats2raw/test/ZarrTest.java
@@ -869,6 +869,27 @@ public class ZarrTest {
   }
 
   /**
+   * Plate defined with no linked images.
+   */
+  @Test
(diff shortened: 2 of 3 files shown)


melissalinkert:
After discussion earlier today, a small OME-XML file that gets committed might be the easiest way to test this.

@muhanadz would you be willing to put together a test file?  Two examples for reference:

https://downloads.openmicroscopy.org/images/OME-XML/2016-06/single-image.ome.xml
https://downloads.openmicroscopy.org/images/OME-XML/2016-06/hcs.ome.xml

It basically needs to have one ```Image``` defined as in the first example, plus an empty ```Plate``` element.  Running ```bioformats2raw test.ome.xml test.zarr``` on the finished test file without this PR should result in weird behavior where the HCS layout (https://ngff.openmicroscopy.org/latest/#hcs-layout) is used but one or more directories are named ```-1```.  The same test with this PR should result in the normal non-HCS layout.

Once a file is ready to go, we can add a corresponding test here.

muhanadz:
Without PR the following section of the XML gets processed and outputs a Zarr file with an HCS layout where one of the directories is named `-1`. 
```
./bioformats2raw-0.3.0/bin/bioformats2raw test_bioformats2raw_pr118.ome.xml test_withoutPR.zarr --compression=zlib
```
```
  <Plate
     ID="Plate:1"
     Name="Control Plate"
     ColumnNamingConvention="letter"
     RowNamingConvention="number"
     Columns="12"
     Rows="8"
     >
  </Plate>
```


With PR the following warning message is displayed and the resulting Zarr is a non-HCS layout (like `--no-hcs`) with normal namings:
```
./bioformats2raw-0.3.1-SNAPSHOT/bin/bioformats2raw test_bioformats2raw_pr118.ome.xml test_withPR.zarr --compression=zlib 
2021-11-08 11:17:52,815 [main] WARN  c.g.bioformats2raw.Converter - Encountered invalid plate #0
```

Adding more than one empty plate doesn't change the documented behaviour, but only displays `Encountered invalid plate #` for the number of empty plates. Removing the defined image in the XML results in `Invalid series: 0` which is expected with PR and without PR

[test_withoutPR_no-hcs.zarr.zip](https://github.com/glencoesoftware/bioformats2raw/files/7504282/test_withoutPR_no-hcs.zarr.zip)
[test_withoutPR.zarr.zip](https://github.com/glencoesoftware/bioformats2raw/files/7504283/test_withoutPR.zarr.zip)
[test_withPR.zarr.zip](https://github.com/glencoesoftware/bioformats2raw/files/7504284/test_withPR.zarr.zip)
[test_bioformats2raw_pr118.ome.xml.zip](https://github.com/glencoesoftware/bioformats2raw/files/7504285/test_bioformats2raw_pr118.ome.xml.zip)

muhanadz (approved):
From previous review, fixed test.ome.xml file to pass  bftools' `xmlvalid` as it wasn't before
```
bftools_v6.5.1-397/xmlvalid ./test_bioformats2raw_pr118.ome.xml 
Parsing schema path
http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd
Validating ./test_bioformats2raw_pr118.ome.xml
No validation errors found.
```
The behaviour with and without the PR remains the same as previously documented with this test.ome.xml file.


[test_bioformats2raw_pr118.ome.xml.zip](https://github.com/glencoesoftware/bioformats2raw/files/7506424/test_bioformats2raw_pr118.ome.xml.zip)
[test_withoutPR.zarr.zip](https://github.com/glencoesoftware/bioformats2raw/files/7506425/test_withoutPR.zarr.zip)
[test_withPR.zarr.zip](https://github.com/glencoesoftware/bioformats2raw/files/7506426/test_withPR.zarr.zip)

melissalinkert:
That's perfect, thanks @muhanadz.  52d1de0 commits the file and adds the corresponding unit test. /cc @chris-allan 

I of course completely forgot that we had already added some small OME-XML tests in #91, so adding another one was easier than expected.

muhanadz (approved):
`gradle build` on this PR builds fine with no problems.
```
BUILD SUCCESSFUL in 1m 55s
11 actionable tasks: 11 executed
```

PR #123, opened 2021-12-23
Feature/improve docker image

ilveroluca:
This PR revises the Dockerfile generate a lighter yet complete Docker image that follows common entrypoint/cmd conventions.   

Image size is reduced by using a two-stage build, avoiding the installation of packages that don't seem to be required, and cleaning up temporary installation files.  The final image is based on the openjdk:8 Docker image rather than the gradle image.  The gradle version used for the build is bumped to 6.9 (seems to me that's the oldest still supported version).

The image has been tested converting a number of Mirax and SVS slides to raw format.

Changing the entrypoint changes the way one starts a container with the image.  Namely, the command `bioformats2raw` needs to be specified in the command line, i.e.:

    docker run -it --rm ilveroluca/bioformats2raw bioformats2raw bioformats2raw --log-level=INFO input.mrxs output.raw

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ ARG BUILD_IMAGE=gradle:6.9-jdk8
 #
 FROM ${BUILD_IMAGE} as build
 USER root
-RUN apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get install -y -qq zeroc-ice-all-runtime libblosc1
+RUN apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get install -y -qq libzeroc-ice3.7-java libblosc1
 RUN mkdir /bioformats2raw && chown 1000:1000 /bioformats2raw
 
 # Build all
@@ -26,8 +26,20 @@ COPY --chown=1000:1000 . /bioformats2raw
 WORKDIR /bioformats2raw
 RUN gradle build
 RUN cd build/distributions && rm bioformats2raw*tar && unzip bioformats2raw*zip && rm -rf bioformats2raw*zip
+
+
+
+FROM openjdk:8 as final
+
+RUN DEBIAN_FRONTEND=noninteractive \
+    apt-get update -y -q \
+ && apt-get install -y -q libzeroc-ice3.7-java libblosc1 \
+ && apt-get autoremove -y && apt-get clean -y && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+
 USER root
-RUN mv /bioformats2raw/build/distributions/bioformats2raw* /opt/bioformats2raw
+COPY --from=build /bioformats2raw/build/distributions/bioformats2raw* /opt/bioformats2raw
+
 USER 1000
 WORKDIR /opt/bioformats2raw
 ENTRYPOINT ["/opt/bioformats2raw/bin/bioformats2raw"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,4 +42,4 @@ ENV PATH="/opt/bioformats2raw/bin:${PATH}"
 
 USER 1000
 WORKDIR /opt/bioformats2raw
-CMD ["bioformats2raw"]
+ENTRYPOINT ["/opt/bioformats2raw/bin/bioformats2raw"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ ARG BUILD_IMAGE=gradle:6.9-jdk8
 #
 FROM ${BUILD_IMAGE} as build
 USER root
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -9,14 +9,14 @@
...
-ARG BUILD_IMAGE=gradle:6.2.1-jdk8
+ARG BUILD_IMAGE=gradle:6.9-jdk8
 
 #
 # Build phase: Use the gradle image for building.
 #
 FROM ${BUILD_IMAGE} as build
 USER root
-RUN apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get install -y -qq zeroc-ice-all-runtime libblosc1
+RUN apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get install -y -qq libzeroc-ice3.7-java libblosc1
```
joshmoore, line 19: I'm very much struggling to remember why zeroc would be needed here. It's almost certainly a copy-n-paste error and removing it works for me locally.

joshmoore, line 19:
```suggestion
RUN apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get install -y -qq libblosc1
```

Dockerfile
```diff
@@ -26,8 +26,20 @@ COPY --chown=1000:1000 . /bioformats2raw
 WORKDIR /bioformats2raw
 RUN gradle build
 RUN cd build/distributions && rm bioformats2raw*tar && unzip bioformats2raw*zip && rm -rf bioformats2raw*zip
+
+
+FROM openjdk:8 as final
+
+RUN DEBIAN_FRONTEND=noninteractive \
+    apt-get update -y -q \
+ && apt-get install -y --no-install-recommends -q libzeroc-ice3.7-java libblosc1 \
```
joshmoore, line 35:
```suggestion
 && apt-get install -y --no-install-recommends -q libblosc1 \
```

Dockerfile
```diff
@@ -26,8 +26,20 @@ COPY --chown=1000:1000 . /bioformats2raw
...
+
 USER root
-RUN mv /bioformats2raw/build/distributions/bioformats2raw* /opt/bioformats2raw
+COPY --from=build /bioformats2raw/build/distributions/bioformats2raw* /opt/bioformats2raw
+ENV PATH="/opt/bioformats2raw/bin:${PATH}"
+
 USER 1000
 WORKDIR /opt/bioformats2raw
-ENTRYPOINT ["/opt/bioformats2raw/bin/bioformats2raw"]
+CMD ["bioformats2raw"]
```
joshmoore, line 45: @ilveroluca, what's the benefit of CMD over ENTRYPOINT here? With CMD,

```
docker run -ti --rm josh-bf2raw-pr123 -h
```
fails.

joshmoore (commented):
Thanks for the slimification & modernization here, @ilveroluca! A few minor suggestions inline.

Dockerfile
ilveroluca: Indeed, because this image does not specify an entrypoint on needs to specify the entire command line when launching the container, i.e.:
```
docker run -ti --rm josh-bf2raw-pr123 bioformats2raw -h
```

ENTRYPOINT vs CMD is often a matter of debate.  The main benefit of not specifying the entrypoint is in terms of flexibility.  The image becomes more easily usable as an environment in which to run bioformats2raw and also other related commands (naturally, this could also be done by overriding the entrypoint...).  However, it does come at the cost of usability.

The concrete use case that led me to make this change is actually gone now.  I needed to run something like
```
mkdir -p /multi/level/output/path/ && bioformats2raw ... /multi/level/output/path/slide.raw
```
in an image conversion workflow I wrote with Snakemake, and being able to run the whole command within the same container made things a bit simpler for the workflow.  However, I have since learned that Snakemake uses container images (including Docker images) by opening a `singularity shell` and thus completely ignores both ENTRYPOINT and CMD.

At this point, I would be inclined to restore the original ENTRYPOINT for simplicity.

joshmoore: :+1:

joshmoore:
Thanks, @ilveroluca. Appveyor failure is unrelated ("Read timed out")

chris-allan:
Thanks for the contribution, @ilveroluca!

ilveroluca:
Thanks for the software :-)

Since you have an organization on DockerHub, it would be good to use the dockerfile to have official images pushed up there! :-)

chris-allan:
Thanks for the feedback, Luca. The dockerfile and the Docker based workflow was really put together by @joshmoore and the OME team to satisfy a series of things they were doing. It's not something that @glencoesoftware really wants to support officially at this time.

PR #134, opened 2022-03-01
First try at scale transforms

melissalinkert:
See https://github.com/ome/ngff/pull/57, follow-up to #121.

The main idea is to preserve physical sizes in the Zarr metadata itself, so that downstream applications don't have to read those values from the OME-XML. There is an open question of what to do for subresolutions; OME-XML won't store any physical sizes for those, so we could omit or store a calculated value.

/cc @sbesson @dgault (since I can't add external reviewers)

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -19,7 +19,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
@@ -56,6 +55,7 @@ import loci.formats.meta.IMetadata;
 import loci.formats.ome.OMEXMLMetadata;
 import loci.formats.services.OMEXMLService;
 import loci.formats.services.OMEXMLServiceImpl;
+import ome.units.quantity.Quantity;
 import ome.xml.meta.OMEXMLMetadataRoot;
 import ome.xml.model.enums.DimensionOrder;
 import ome.xml.model.enums.EnumerationException;
@@ -1506,33 +1506,63 @@ public class Converter implements Callable<Void> {
     multiscale.put("metadata", metadata);
     multiscale.put("version", nested ? "0.2" : "0.1");
     multiscales.add(multiscale);
-    List<Map<String, String>> datasets = new ArrayList<Map<String, String>>();
+
+    IFormatReader v = null;
+    IMetadata meta = null;
+    String axisOrder = null;
+    try {
+      v = readers.take();
+      meta = (IMetadata) v.getMetadataStore();
+
+      if (dimensionOrder != null) {
+        axisOrder = dimensionOrder.toString();
+      }
+      else {
+        axisOrder = v.getDimensionOrder();
+      }
+    }
+    finally {
+      readers.put(v);
+    }
+
+    List<Map<String, Object>> datasets = new ArrayList<Map<String, Object>>();
     for (int r = 0; r < resolutions; r++) {
       resolutionString = String.format(
               scaleFormatString, getScaleFormatStringArgs(series, r));
       String lastPath = resolutionString.substring(
               resolutionString.lastIndexOf('/') + 1);
-      datasets.add(Collections.singletonMap("path", lastPath));
+
+      List<Map<String, Object>> transforms =
+        new ArrayList<Map<String, Object>>();
+      Map<String, Object> scale = new HashMap<String, Object>();
+      scale.put("type", "scale");
+      List<Double> axisValues = new ArrayList<Double>();
+      List<Integer> axisIndices = new ArrayList<Integer>();
+      for (int i=axisOrder.length()-1; i>=0; i--) {
+        Quantity axisScale = getScale(meta, series, axisOrder, i);
+        if (axisScale != null) {
+          // TODO: does this need to be recalculated for r > 0?
+          axisValues.add(axisScale.value().doubleValue());
+          axisIndices.add(axisOrder.length() - i - 1);
+        }
+      }
+      scale.put("scale", axisValues);
+      scale.put("axisIndices", axisIndices);
+
+      transforms.add(scale);
+
+      Map<String, Object> dataset = new HashMap<String, Object>();
+      dataset.put("path", lastPath);
+      dataset.put("transformations", transforms);
+      datasets.add(dataset);
     }
     multiscale.put("datasets", datasets);
 
-    String axisOrder = null;
-    if (dimensionOrder != null) {
-      axisOrder = dimensionOrder.toString();
-    }
-    else {
-      IFormatReader reader = readers.take();
-      try {
-        axisOrder = reader.getDimensionOrder();
-      }
-      finally {
-        readers.put(reader);
-      }
-    }
     List<Map<String, String>> axes = new ArrayList<Map<String, String>>();
     for (int i=axisOrder.length()-1; i>=0; i--) {
       String axis = axisOrder.substring(i, i + 1).toLowerCase();
       String type = "space";
+      Quantity scale = getScale(meta, series, axisOrder, i);
       if (axis.equals("t")) {
         type = "time";
       }
@@ -1542,6 +1572,9 @@ public class Converter implements Callable<Void> {
       Map<String, String> thisAxis = new HashMap<String, String>();
       thisAxis.put("name", axis);
       thisAxis.put("type", type);
+      if (scale != null) {
+        thisAxis.put("unit", scale.unit().getSymbol());
+      }
       axes.add(thisAxis);
     }
     multiscale.put("axes", axes);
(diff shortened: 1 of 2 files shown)


sbesson (changes requested):
Some feedback based on the initial conversion of a multi-image CZI files. In general, completely agreed on limiting this implementation to mapping `OME.PixelSize{X,Y,Z}` into `scale` as this is the minimal transformation allowing compliance with the 0.4 specification.

Looking at the content of the `.zattrs` for one of the multiscales images:

```json
{
  "multiscales" : [
    {
      "metadata" : {
        "version" : "Bio-Formats 6.8.0"
      },
      "axes" : [
        {
          "name" : "t",
          "type" : "time"
        },
        {
          "name" : "c",
          "type" : "channel"
        },
        {
          "name" : "z",
          "type" : "space"
        },
        {
          "unit" : "µm",
          "name" : "y",
          "type" : "space"
        },
        {
          "unit" : "µm",
          "name" : "x",
          "type" : "space"
        }
      ],
      "datasets" : [
        {
          "path" : "0",
          "transformations" : [
            {
              "scale" : [
                0.21998680079195246,
                0.21998680079195246
              ],
              "type" : "scale",
              "axisIndices" : [
                3,
                4
              ]
            }
          ]
        },
        {
          "path" : "1",
          "transformations" : [
            {
              "scale" : [
                0.21998680079195246,
                0.21998680079195246
              ],
              "type" : "scale",
              "axisIndices" : [
                3,
                4
              ]
            }
          ]
        },
        {
          "path" : "2",
          "transformations" : [
            {
              "scale" : [
                0.21998680079195246,
                0.21998680079195246
              ],
              "type" : "scale",
              "axisIndices" : [
                3,
                4
              ]
            }
          ]
        },
...
```

- `axisIndices` has been reverted in the latest push for the 0.4 `coordinateTransformations` specification in favor of a simplified `scale`, `translate` vectore - see https://github.com/ome/ngff/pull/85/files. Unless the goal
- `scale` should include all dimensions and use 1.0 when there is not metadata mapping to the physical space
- unlike in the case of Bio-Formats where we read native sub-resolutions, my understanding is that the `Converter`  always generates the pyramidal levels with a known downsampling factor. In that case, I think the `scale` vector should be computed for each `dataset` using the original pixel size and the downsampling factor.

melissalinkert:
I think d6fcc38 addresses everything in https://github.com/glencoesoftware/bioformats2raw/pull/134#pullrequestreview-897606775.

sbesson:
With the last commit, running a conversion on a real pyramidal input file leads to the expected layout

```json
{
  "multiscales" : [
    {
      "metadata" : {
        "version" : "Bio-Formats 6.8.0"
      },
      "axes" : [
        {
          "name" : "t",
          "type" : "time"
        },
        {
          "name" : "c",
          "type" : "channel"
        },
        {
          "name" : "z",
          "type" : "space"
        },
        {
          "unit" : "µm",
          "name" : "y",
          "type" : "space"
        },
        {
          "unit" : "µm",
          "name" : "x",
          "type" : "space"
        }
      ],
      "datasets" : [
        {
          "path" : "0",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                0.21998680079195246,
                0.21998680079195246
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "1",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                0.4399736015839049,
                0.4399736015839049
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "2",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                0.8799472031678098,
                0.8799472031678098
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "3",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                1.7598944063356197,
                1.7598944063356197
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "4",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                3.5197888126712393,
                3.5197888126712393
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "5",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                7.039577625342479,
                7.039577625342479
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "6",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                14.079155250684957,
                14.079155250684957
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "7",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                28.158310501369915,
                28.158310501369915
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "8",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                56.31662100273983,
                56.31662100273983
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "9",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                112.63324200547966,
                112.63324200547966
              ],
              "type" : "scale"
            }
          ]
        }
      ],
      "version" : "0.2"
    }
  ]
}
```

The unit test expansion should also cover the testing of the scale values.
Also testing the conversion of a minimal `test.fake` without any physical size produces a compliant output

```json
{
  "multiscales" : [
    {
      "metadata" : {
        "version" : "Bio-Formats 6.8.0"
      },
      "axes" : [
        {
          "name" : "t",
          "type" : "time"
        },
        {
          "name" : "c",
          "type" : "channel"
        },
        {
          "name" : "z",
          "type" : "space"
        },
        {
          "name" : "y",
          "type" : "space"
        },
        {
          "name" : "x",
          "type" : "space"
        }
      ],
      "datasets" : [
        {
          "path" : "0",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                1.0,
                1.0
              ],
              "type" : "scale"
            }
          ]
        },
        {
          "path" : "1",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                1.0,
                1.0
              ],
              "type" : "scale"
            }
          ]
        }
      ],
      "version" : "0.2"
    }
  ]
}
```

This example raised a question related to the following sentence in the 0.4 specification "If scaling information is not available or applicable for one of the axes set it to 1. ". Given the absence of physical metadata, I would consider that the scale vectors used above are in agreement with this statement.
However, one could certainly conceive that the resolution of all resolution levels beyond the first one should include the downscaling factor i.e. the expectation should be:

```json
        {
          "path" : "1",
          "coordinateTransformations" : [
            {
              "scale" : [
                1.0,
                1.0,
                1.0,
                2.0,
                2.0
              ],
              "type" : "scale"
            }
``` 

@constantinpape is there an authoritative answer to this concern? Depending on the outcome, happy to open a PR or an issue against the ome/ngff related to this clarification.

Once this is settled, the changes proposed here all look good to me and should allow us to bump the `multiscales` version as proposed in #135.

constantinpape:
> @constantinpape is there an authoritative answer to this concern? Depending on the outcome, happy to open a PR or an issue against the ome/ngff related to this clarification.

Good catch; in this case we still need to add scaling information for the axes that are actually downscaled. So the second snippet for dataset `"1"` would be correct. (Otherwise the spaces wouldn't align). It would be good if you can open an issue / PR about this in ngff.

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -1506,33 +1506,63 @@ private void setSeriesLevelMetadata(int series, int resolutions)
...
-        readers.put(reader);
+      List<Map<String, Object>> transforms =
+        new ArrayList<Map<String, Object>>();
+      Map<String, Object> scale = new HashMap<String, Object>();
+      scale.put("type", "scale");
+      List<Double> axisValues = new ArrayList<Double>();
+      double resolutionScale = Math.pow(PYRAMID_SCALE, r);
+      for (int i=axisOrder.length()-1; i>=0; i--) {
+        Quantity axisScale = getScale(meta, series, axisOrder, i);
+        if (axisScale != null) {
```
sbesson, line 1543: This will need to be updated to store the following values:
- `axisScale.value().doubleValue() * resolutionScale` if downsampling happens along the dimension and there is some scaling metadata associated with the dimension
- `resolutionScale` if downsampling happens along the dimension and there is no scaling metadata associated with the dimension
- `1.0` otherwise

sbesson (changes requested):
https://ngff.openmicroscopy.org/0.4/ has been updated to clarify the fact that the `scale` vector should at minimum contain information about the downsampling of each resolution i.e. in the case of data without physical sizes downsampled by a factor of 2 along X and Y

`[1.0, 1.0, 1.0, 1.0, 1.0]` for the first resolution level
`[1.0, 1.0, 1.0, 2.0, 2.0]` for the second resolution level
`[1.0, 1.0, 1.0, 4.0, 4.0]` for the third resolution level
...

sbesson (approved):
Tested the last commit with several real and fake datasets allowing to test combinations of output. The transformation metadata is stored as expected including in the case where no physical metadata exists.

The expanded and new unit test should cover the different scenarios (dimension with scaling information and downsampling, dimension with no scaling information and downsampling, dimension with scaling information and no downsampling, dimension with no scaling information and no dowsampling).

Good to merge from my side

PR #135, opened 2022-03-02
Version bumps

melissalinkert:
/cc @sbesson

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,7 @@ configurations.all {
 }
 
 dependencies {
-    implementation 'ome:formats-gpl:6.8.0'
+    implementation 'ome:formats-gpl:6.9.0'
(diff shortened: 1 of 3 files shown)


sbesson:
From my side, :+1: for bumping Bio-Formats immediately.

On the OME-NGFF version bump, I think it depends on the strategy and the expected timeline for #134 as the latter should be the last step to bump the `multiscales` layout to `0.4`. Using `0.3` would be an possible intermediate step but in case this was meant to be released, additional changes would need to be made on top of #121 anyways as the `axes` specification changed between 0.3 and 0.4 from a string to a dictionary.

melissalinkert:
#134 is meant to be merged first, then this one as a last step before a new release. I opened this now so that I don't forget to do it later.

chris-allan (approved):

PR #143, opened 2022-03-29
MRXS: update how 2 channel fluorescence JPEG data is handled

melissalinkert:
JPEG-2000 data with 2 channels is known to have the third stored channel as filler.
What little 2 channel JPEG data we have uses the first channel as filler instead.


melissalinkert:
Private notes for reviewers: https://glencoesoftware.hackmd.io/Ri2dIW3rT0CXMmn-8Q4kag (/cc @mgheirat)

mgheirat:
Tested on the problematic mrxs data provided in the private notes. The missing nuclear channel is showing correctly along with the second channel. Procedure: Bioformats2raw/raw2ometiff, imported ome-tiff to OMERO (via Insight), viewed in PathViewer. The intermediary zarr is also read programmatically in python and the two channels visualized separately. They looked fine. 
The procedure is tested on an older mrxs file as well (as marked in the private notes) and the resulting zarr and ome-tiff images both looked fine.

Issue #136, opened 2022-03-02
Delete existing memo files by default?


melissalinkert:
If a memo file already exists before the conversion is started, it will be used if it is compatible with the current Bio-Formats version. This can lead to confusion if conversion is being tested with different reader options, since the memo file effectively overrides any options changes.

It might be better to default to deleting the memo file (if it exists) before the first ```setId``` call, and add an option to keep the memo file for the rare cases when that is necessary. ```Memoizer.getMemoFile(String)``` should make that easy.

sbesson:
Another (complementary?) approach would be to delete the memo file at the end of the conversion process.

Since the primary scope of this utility is to perform a one-off conversion and this file is primarily generated for  performance, at least from my side, there is no persistence expectation and it would be completely acceptable to delete it at the end of the process.

joshmoore:
With the caveat that if it exists beforehand, it likely shouldn't be deleted by default afterwards (e.g. in the IDR case)

PR #145, opened 2022-03-31
Delete memo files by default

melissalinkert:
Fixes https://github.com/glencoesoftware/bioformats2raw/issues/136

There are 3 cases covered here:

- `--keep-memo-files` is used; no memo files will be deleted
- memo file is not created, either because a valid memo file existed before bioformats2raw was run or the initialization time was too short; no memo files will be deleted
- memo file was created and `--keep-memo-files` is *not* used; memo file will be deleted at the end of conversion

The test is turned off at the moment as I need to figure out a way to force memo file creation for fake files (see https://github.com/ome/bioformats/pull/3804).

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -298,6 +298,12 @@ public class Converter implements Callable<Void> {
   private volatile File memoDirectory;
 
   @Option(
+          names = "--keep-memo-files",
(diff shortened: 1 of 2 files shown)


sbesson:
I expect the just-released Bio-Formats 6.9.1 artifacts should allow to activate the unit tests

joshmoore:
Other than that this points to improvements for the `Memoizer` API (accept null arguments, allow deleting the memo file, etc.), this logic makes sense to me.

PR #148, opened 2022-04-25
Update jzarr to 0.3.5

melissalinkert:
Switches from GS snapshot to 0.3.5, which is the current latest version.

The API for specifying nested chunks is slightly different, so this isn't just a version bump. While the separator can easily be set when creating an array, there is no API method to get the array's separator after the fact. Tests here now check for the existence of the first chunk file (which has a hard-coded path), but we could do something like what the jzarr tests do instead:

https://github.com/bcdev/jzarr/blob/0.3.5/src/test/java/com/bc/zarr/ZarrArrayTest_dimensionSeparator.java#L94

raw2ometiff will need a similar treatment (started on https://github.com/glencoesoftware/raw2ometiff/compare/master...melissalinkert:jzarr-version?expand=1), but not until this is in a released version of bioformats2raw.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ dependencies {
     implementation 'ome:formats-gpl:6.9.1'
     implementation 'info.picocli:picocli:4.6.1'
     implementation 'com.univocity:univocity-parsers:2.8.4'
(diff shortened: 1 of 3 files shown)


melissalinkert:
See https://github.com/bcdev/jzarr/pull/32

PR #149, opened 2022-04-27
HCS well sample/image mapping

sbesson:
Discovered while investigating https://forum.image.sc/t/ome-zarr-conversion-tool/62234/29

The current logic for assiging well samples to series is based on `WellSample.Index` and the assumption this corresponds to the series index. For most Bio-Formats HCS readers, the reader generates and populates this metadata and this assumption holds.
Looking at the specification, the [WellSample.Index](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#WellSample_Index) is defined more liberally as there is no clear expectation is should map to the `Image` order, indices do not need to start at `0` and gaps are allowed. Examples of such indexing can be found in the official OME-XML/OME-TIFF samples like the [plate-companion OME-TIFF](https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/plate-companion/hcs.companion.ome)  or [one-screen-one-plate-four-wells OME-XML](https://downloads.openmicroscopy.org/images/OME-XML/2016-06/one-screen-one-plate-four-wells.ome.xml) samples.

With bioformats2raw 0.4.0, converting both samples will create a `-1/-1/-1` Zarr hierarchy as described in the thread above. Using the current `HEAD` of this repository, the converter will error  while trying to assign a series to a well sample in https://github.com/glencoesoftware/bioformats2raw/blob/8eb4aae1017e5c5a642a0791e8d0ae60d1d81bed/src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java#L57-L58.

878a586adff3d08c0eebff30b43d3f82d808e4de proposes to solve this issue by replacing the index-based check by a reference based check using `WellSample.ImageRef` and `Image.ID`.
7297dd2434d4b4c00ddf97fe71c999535730bf2c fixes a secondary issue that arises if multiple `PlateAcquisition` elements are defined.

The two OME-XML and OME-TIFF samples mentioned above can be used to test this PR:
- the conversion command should complete successfully
- the conversion should contain no more Zarr groups named `-1`
- for `one-screen-one-plate-four-wells.ome.xml`, two `acquisitions` elements should be defined in the top-level `.zattrs` and be referenced in the well attributes

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java
@@ -29,7 +29,7 @@ public class HCSIndex {
     for (int p=0; p<meta.getPlateCount(); p++) {
       for (int w=0; w<meta.getWellCount(p); w++) {
         for (int ws=0; ws<meta.getWellSampleCount(p, w); ws++) {
-          if (meta.getWellSampleIndex(p, w, ws).getValue() == series) {
+          if (meta.getWellSampleImageRef(p, w, ws) == meta.getImageID(series)) {
             plate = p;
             field = ws;
             wellRow = meta.getWellRow(p, w).getValue();
--- a/src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java
@@ -44,7 +44,7 @@ public class HCSIndex {
                   break;
                 }
               }
-              if (thePlateAcquisition >= 0) {
+              if (thePlateAcquisition != null) {
                 break;
               }
             }
--- a/src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java
@@ -29,7 +29,9 @@ public class HCSIndex {
     for (int p=0; p<meta.getPlateCount(); p++) {
       for (int w=0; w<meta.getWellCount(p); w++) {
         for (int ws=0; ws<meta.getWellSampleCount(p, w); ws++) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java
```diff
@@ -29,7 +29,7 @@ public HCSIndex(IMetadata meta, int series) {
     for (int p=0; p<meta.getPlateCount(); p++) {
       for (int w=0; w<meta.getWellCount(p); w++) {
         for (int ws=0; ws<meta.getWellSampleCount(p, w); ws++) {
-          if (meta.getWellSampleIndex(p, w, ws).getValue() == series) {
+          if (meta.getWellSampleImageRef(p, w, ws) == meta.getImageID(series)) {
```
melissalinkert, line 32: This comparison should use `.equals` instead of `==`.

melissalinkert (changes requested):
I think this approach makes sense based on discussions from earlier today. It's probably worth adding a unit test that would have caught this issue, but I can do that separately if you prefer.

src/main/java/com/glencoesoftware/bioformats2raw/HCSIndex.java
sbesson: Should be fixed in 029df918a09fa61a9db9d9c161926e6b389babc6

sbesson:
[d46487d](https://github.com/glencoesoftware/bioformats2raw/pull/149/commits/d46487dead72654d56ca77f75b6c74788d88341d) adds a unit test which should cover both scenarios adressed in this PR: a well with 2 plate acquisitions and wellsample indices not matching the series indices.

melissalinkert (approved):

PR #151, opened 2022-04-29
Add info in installing via conda

will-moore:
Didn't know this option existed till @sbesson mentioned it. 
Probably useful for others too.


melissalinkert:
Thanks @will-moore. I think it would be better to have this just point to https://github.com/ome/conda-bioformats2raw, and put specific installation instructions in that repo's readme.

PR #152, opened 2022-05-05
Fix the type for acquisitions.id under the plate metadata

sbesson:
Although the type is not specified in https://ngff.openmicroscopy.org/0.4/#plate-md, the snippets use integers for the aqcuisitions identifiers. The `well.acquisition` metadata is also of integer type to create a reference to the acquisitions objects defined at the `plate` level.

39237e24b4eb08f73b13d75c9d416cc3bfdd8201 updates the current writer code and adjusts the tests accordingly. Note this PR is currently built on top of #149 which adds a new HCS test with the previous behavior for the `id` type.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -1381,7 +1381,7 @@ public class Converter implements Callable<Void> {
       new ArrayList<Map<String, Object>>();
     for (int pa=0; pa<meta.getPlateAcquisitionCount(plate); pa++) {
       Map<String, Object> acquisition = new HashMap<String, Object>();
-      acquisition.put("id", String.valueOf(pa));
+      acquisition.put("id", pa);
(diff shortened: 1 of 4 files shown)


melissalinkert (approved):
I'm fine with this, but it probably makes sense to update the spec as well to clarify the correct type.

melissalinkert:
I'd expect data generated by current `HEAD` of bioformats2raw to work with current `HEAD` of raw2ometiff. I wouldn't try to update the bioformats2raw dependency in raw2ometiff locally (due to jzarr version differences).

Keep in mind though that https://github.com/glencoesoftware/raw2ometiff/blob/master/src/main/java/com/glencoesoftware/pyramid/PyramidFromDirectoryWriter.java#L376 will only be executed if there is no OME-XML metadata file:

https://github.com/glencoesoftware/raw2ometiff/blob/master/src/main/java/com/glencoesoftware/pyramid/PyramidFromDirectoryWriter.java#L624

PR #155, opened 2022-05-16
Add NGFF spec version to `--version` output

melissalinkert:
See https://github.com/ome/omero-cli-zarr/issues/113


sbesson (approved):
Moving the NGFF specification version as a class variable and exposing it in the `version` digest makes sense.

092ae52566108e91e9a3be88c5dd950774036be4 uses it in the `plate` specification with the corresponding test. Happy for this commit to be cherry-picked here or open a follow-up PR.

PR #156, opened 2022-05-18
Add MetadataOnly to METADATA.ome.xml

### LINKED ISSUE #137: Add `MetadataOnly` to `METADATA.ome.xml`?
Discussed briefly in formats meeting (/cc @joshmoore, @sbesson, @dgault). 

`METADATA.ome.xml` doesn't technically validate, because it doesn't have `MetadataOnly`, `BinData`, or `TiffData` under `Pixels`. This was sort of intentional, as adding a `MetadataOnly` here means that raw2ometiff then needs to remove it.  Since we're now using the output of bioformats2raw independent of raw2ometiff and suggesting that other applications could use `METADATA.ome.xml`, it probably makes sense to update this.

melissalinkert:
Fixes #137

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -632,6 +632,9 @@ public class Converter implements Callable<Void> {
         }
 
         if (!noOMEMeta) {
+          for (int s=0; s<meta.getImageCount(); s++) {
+            getService().addMetadataOnly((OMEXMLMetadata) meta, s, s == 0);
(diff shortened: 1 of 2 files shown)


melissalinkert:
Datasets converted with this change should be convertible to OME-TIFF with no changes in raw2ometiff, see https://github.com/glencoesoftware/raw2ometiff/blob/master/src/main/java/com/glencoesoftware/pyramid/PyramidFromDirectoryWriter.java#L617. `tiffcomment | grep MetadataOnly` can confirm that the `MetadataOnly` is already removed.

PR #157, opened 2022-06-04
Add group path to series index mapping in root attributes

### LINKED ISSUE #126: Enumerate series in zarr metadata?
@glencoesoftware team briefly discussed that it might be helpful to include series paths/indexes and names in top-level zarr metadata so that OME-XML doesn't always have to be parsed.

melissalinkert:
Fixes #126.

The root level attributes (if they are written) will now contain a `groups` dictionary that maps each group with multiscales metadata to the corresponding Bio-Formats series index (== Image index in `METADATA.ome.xml`).

As noted in the comments, the paths are dictionary keys and the integer indexes are the values. If the indexes were keys, they would be stored as strings, so this seems easier for a reader. Example tiny plate:

```
$ bin/bioformats2raw "test&plateRows=3&plateCols=2.fake" test-plate
$ head -n 10 test-plate/.zattrs
{
  "groups" : {
    "C/1/0" : 4,
    "C/2/0" : 5,
    "B/1/0" : 2,
    "A/1/0" : 0,
    "B/2/0" : 3,
    "A/2/0" : 1
  },
  "bioformats2raw.layout" : 3,
```

Happy to hear better ideas overall (especially for the attribute name), this is just a place to start.

/cc @kkoz @erindiel @DavidStirling @joshmoore @dgault @sbesson @chris-allan (and feel free to add anyone I missed)

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -665,6 +665,32 @@ public class Converter implements Callable<Void> {
         scaleFormatString = "%s/%s/%d/%d";
       }
 
+      // fileset level metadata
+      if (!noRootGroup) {
+        final ZarrGroup root = ZarrGroup.create(getRootPath());
+        Map<String, Object> attributes = new HashMap<String, Object>();
+        attributes.put("bioformats2raw.layout", LAYOUT);
+
+        // record the path to each series (multiscales) and the corresponding
+        // series (OME-XML Image) index
+        // using the index as the key would mean that the index is stored
+        // as a string instead of an integer
+        Map<String, Integer> groupMap = new HashMap<String, Integer>();
+        for (Integer index : seriesList) {
+          String resolutionString = String.format(
+                  scaleFormatString, getScaleFormatStringArgs(index, 0));
+          String seriesString = "";
+          if (resolutionString.indexOf('/') >= 0) {
+            seriesString = resolutionString.substring(0,
+                resolutionString.lastIndexOf('/'));
+          }
+          groupMap.put(seriesString, index);
+        }
+        attributes.put("groups", groupMap);
+
+        root.writeAttributes(attributes);
+      }
+
       for (Integer index : seriesList) {
         try {
           write(index);
@@ -1214,14 +1240,6 @@ public class Converter implements Callable<Void> {
       "sizeY {} (tileWidth: {}) sizeZ {} (tileDepth: {}) imageCount {}",
         sizeX, tileWidth, sizeY, tileHeight, sizeZ, chunkDepth, imageCount
     );
-
-    // fileset level metadata
-    if (!noRootGroup) {
-      final ZarrGroup root = ZarrGroup.create(getRootPath());
-      Map<String, Object> attributes = new HashMap<String, Object>();
-      attributes.put("bioformats2raw.layout", LAYOUT);
-      root.writeAttributes(attributes);
-    }
 
     // series level metadata
     setSeriesLevelMetadata(series, resolutions);
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -682,7 +682,7 @@ public class Converter implements Callable<Void> {
         // series (OME-XML Image) index
         // using the index as the key would mean that the index is stored
         // as a string instead of an integer
-        Map<String, Integer> groupMap = new HashMap<String, Integer>();
+        List<String> groups = new ArrayList<String>();
         for (Integer index : seriesList) {
           String resolutionString = String.format(
                   scaleFormatString, getScaleFormatStringArgs(index, 0));
@@ -691,9 +691,9 @@ public class Converter implements Callable<Void> {
             seriesString = resolutionString.substring(0,
                 resolutionString.lastIndexOf('/'));
           }
-          groupMap.put(seriesString, index);
+          groups.add(seriesString);
         }
-        attributes.put("series", groupMap);
+        attributes.put("series", groups);
 
         root.writeAttributes(attributes);
       }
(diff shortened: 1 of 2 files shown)


DavidStirling:
This is absolutely amazing!

melissalinkert:
Our use cases aren't specific to HCS, so I'd prefer to have a more generic solution than just extending the `well` specification.

I don't think it makes sense to add more metadata under this new attribute, so a simple dictionary should be enough. If we need more image or well-level metadata, then those specs should be updated accordingly.

In terms of mismatches, I think what is in `.zattrs` should take precedence over the natural ordering of the paths. Thinking of a case like:

```
"groups" : {
    "0" : 1,
    "1" : 2,
    "2" : 0
  },
```

...that likely means a corresponding update to raw2ometiff. If this were written up as part of the spec, it probably means adding a bunch of requirements to the effect of `the path MUST exist`, `the index MUST be greater than 0 and less than the group count`, `each path and index MUST be specified exactly once`, etc. That doesn't address what happens if the mapping is fundamentally wrong (with dimension mismatches), but we arguably have that problem already if the OME-XML is edited incorrectly.

One thought for handling the JSON size issue is to move this out of root `.zattrs` and into a new `OME/.zattrs`. That reinforces that this metadata is tied to `METADATA.ome.xml`, and means that anyone who doesn't need it doesn't have to read it. It's still ~150KB for a plate with 10000 images, so if that's too big we'll need to think about other options.

chris-allan:
Given the current hard mapping between Bio-Formats series and the path to the group which contains the `multiscales` metadata it may make more sense to embrace this rather than attempt to make something generic that isn't. Something like (for plate data):

```
"series": [
  "C/1/0",
  "C/2/0"
]
```

and for non-plate data:

```
"series": [
  "0",
  "1"
]
```

This does what it's supposed to do and no more.

sbesson:
If extensibility is unlikely to be a requirement, given that the series are always indexed from 0 to n, a simple list as proposed in https://github.com/glencoesoftware/bioformats2raw/pull/157#issuecomment-1148431700 would certainly do the job.

> One thought for handling the JSON size issue is to move this out of root .zattrs and into a new OME/.zattrs. That reinforces that this metadata is tied to METADATA.ome.xml, and means that anyone who doesn't need it doesn't have to read it. It's still ~150KB for a plate with 10000 images, so if that's too big we'll need to think about other options.

Interesting. At least from my perspective, this would feel like a good compromise allowing to store this metadata without inflating the top-level `.zattrs`.
Assuming we are still in brainstorming mode, the other option I could think of would be to store such information natively as an [object array](https://zarr.readthedocs.io/en/stable/tutorial.html#object-arrays). I assume the tradeoff here would be the added complexity for accessing the indexing array as opposing to just decoding a JSON vs the extra features associated with using the native Zarr array storage (chunking, compression).

melissalinkert:
beb41fa and 508c9bb together should address https://github.com/glencoesoftware/bioformats2raw/pull/157#issuecomment-1148431700 and move the new attribute to `OME/.zattrs`. It took me a bit to convince myself that this still works with `--series`, but it does because `--series` will remove any unconverted `Images` before writing `METADATA.ome.xml`.

sbesson:
Using the latest commit, I successfully converted a [public IDR plate](https://idr.openmicroscopy.org/webclient/?show=plate-7361) with 384 wells, 30 fields of view i.e. ~10K images. I uploaded the  output to a temporary public bucket https://uk1s3.embassy.ebi.ac.uk/bf2raw_157/7361.zarr in case the Glencoe team already has some tools where to test the series mapping proposal.

melissalinkert:
:+1: thanks @sbesson!

joshmoore:
Very sorry for my long silence on this! 😷 

My highest level comment is whether or not there's interest in trying to move this discussion to ome/ngff as a new spec. To me this feels very much like the first steps towards the non-transitional (permanent?) solution for https://github.com/ome/ngff/issues/104

> [melissalinkert](https://github.com/melissalinkert) commented [14 days ago](https://github.com/glencoesoftware/bioformats2raw/pull/157#issuecomment-1147652576)
> One thought for handling the JSON size issue is to move this out of root .zattrs and into a new OME/.zattrs. That reinforces that this metadata is tied to METADATA.ome.xml, and means that anyone who doesn't need it doesn't have to read it. It's still ~150KB for a plate with 10000 images, so if that's too big we'll need to think about other options.

It might be worth a round of discussions on it, but it seems like the feeling is generally "everything can't be in one JSON file". One thing I might add though is that at the top level we might want to "register" this location. (See example below)

> [chris-allan](https://github.com/chris-allan) commented [13 days ago](https://github.com/glencoesoftware/bioformats2raw/pull/157#issuecomment-1148431700)
> Given the current hard mapping between Bio-Formats series and the path to the group which contains the multiscales metadata it may make more sense to embrace this rather than attempt to make something generic that isn't.

In my first quick attempts at this, I had also sided on the use of an array. I was almost sure I had posted this somewhere, but not being able to find it I'll repeat myself here:

```json
{
    "@type": "GenericContainer",
    "metadataSources": [
        {
            "@type": "OMEXMLMetadata",
            "file": {
                "path": "/OME/METADATA.ome.xml"
            },
            "seriesMapping": [
                "s0", "s1", "s2"
            ]
        }
    ]
}
```

melissalinkert:
https://github.com/melissalinkert/ngff/commits/group-list (on top of https://github.com/ome/ngff/pull/112) is a start at describing what this PR does now.

joshmoore:
Used this current branch to build 2 samples which have been uploaded to uk1s3 for testing. (See https://github.com/ome/ome-ngff-validator/pull/13 for URLs)

sbesson (approved):
I am planning to give https://github.com/ome/ngff/pull/112 another round of review today. My understanding is that the next step there will be to port this addition to the 0.4 specification document and update https://ngff.openmicroscopy.org/0.4/ to clarify the expectations and conventions of this layout for data consumers

My vote is to get this change merged and move towards the release `bioformats2raw 0.5.0`. Once the upstream OME-NGFF specification is updated, the README can also be amended to include a reference to the relevant section.

PR #167, opened 2022-10-17
Explicitly declare logback as library dependencies

sbesson:
Set the version to 1.3.4 which is the latest stable release with JDK 8 support


melissalinkert (approved):
`bioformats2raw -p` and `bioformats2raw --debug DEBUG` behave as expected. `build/distributions/bioformats2raw-0.6.0-SNAPSHOT/lib/` has logback 1.3.4 but no other versions.

PR #169, opened 2022-10-18
Remove logback.xml from jar

melissalinkert:
Fixes #168, I think. Once we're happy with this, raw2ometiff can get the same treatment.

37cd1c3 moves `logback.xml` to `src/dist/lib/config` so that it won't be automatically included in the jar, but will be automatically included in the distribution's `lib` directory. The scripts' classpath is updated to include `src/dist/lib/config`, otherwise `logback.xml` won't end up on the classpath.

6427c4f is just to make testing easier, and can be reverted.

Basic testing:

```
$ ./gradlew clean build
$ cd build/distributions
$ unzip bioformats2raw-0.6.0-SNAPSHOT.zip
$ cd bioformats2raw-0.6.0-SNAPSHOT
$ find . -name logback.xml
./lib/config/logback.xml
$ jar tf lib/bioformats2raw-0.6.0-SNAPSHOT.jar | grep logback
$ bin/bioformats2raw --debug DEBUG test.fake test
# note "bf2raw" prefix on every line
```

PR #171, opened 2022-10-31
Set default channel color to grey for single channel images

### LINKED ISSUE #170: Color metadata & single-channel use case
https://github.com/glencoesoftware/bioformats2raw/pull/160 introduced the population of the [omero](https://ngff.openmicroscopy.org/0.4/#omero-md) metadata specification by this conversion tool.

The new logic populates the mininum/maximum values per channel but also stores some channel metadata like the color. The relevant portion of the code is largely inherited from `omero-blitz` [ChannelProcessor](https://github.com/ome/omero-blitz/blob/5d27e4771af5c4c457464090edb5b450d1c11e1d/src/main/java/ome/formats/model/ChannelProcessor.java#L129) to try to set a meaningful value based on various metadata fields and invoked in

https://github.com/glencoesoftware/bioformats2raw/blob/17910912043ce812a37f006f20ffc971ade7c694/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java#L1788

While testing different conversion workflows, I noticed one particular deviation in the case of single-channel image. This can be tested by

```
% bioformats2raw test.fake test.zarr
% grep color test.zarr/0/.zattrs 
      "color" : "FF0000",
```

In the case of single-channel data, the OMERO importer calls a separate `setSingleChannel` method which sets the color to grey instead.

https://github.com/ome/omero-blitz/blob/5d27e4771af5c4c457464090edb5b450d1c11e1d/src/main/java/ome/formats/model/ChannelProcessor.java#L106-L119

Should `bioformats2raw` be updated to adopt the same behavior?

melissalinkert:
Fixes #170

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Colors.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Colors.java
@@ -35,6 +35,7 @@ public class Colors {
   /**
    * Determine which color to use for the given channel.
    * If a channel color is already set, that will be returned.
+   * If there is only one channel, it will be set to grey.
    * If a channel emission wavelength is set, that will be used
    * to determine the color; red, green, or blue based on the
    * wavelength in nm.
@@ -52,6 +53,10 @@ public class Colors {
     Color color = meta.getChannelColor(series, c);
     if (color != null) {
       return color;
+    }
+    // default to grey for single channel data
+    if (meta.getChannelCount(series) == 1) {
+      return new Color(128, 128, 128, 255);
     }
 
     Length emWave = meta.getChannelEmissionWavelength(series, c);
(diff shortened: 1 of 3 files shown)


sbesson (approved):
Tested by converting a minimal test.fake with bioformats2raw 0.5.0 and with this PR included. The only metadata difference is the channel color which is now grey rather than red

```
(java11) sbesson@Sebastiens-MacBook-Pro-2 bioformats2raw % diff test.zarr/0/.zattrs ~/Downloads/test.zarr/0/.zattrs 
41c41
<       "color" : "808080",
---
>       "color" : "FF0000",
(java11) sbesson@Sebastiens-MacBook-Pro-2 bioformats2raw % diff test.zarr/.zattrs ~/Downloads/test.zarr/.zattrs 
(java11) sbesson@Sebastiens-MacBook-Pro-2 bioformats2raw % diff test.zarr/OME/.zattrs test.zarr/OME/.zattrs 
```

This is inline with the current behavior of the OMERO import logic. The extension of the color unit test should cover the most common combinations including single-channel with/without color metadata and multi-channel.

From my side this should be amenable to an immediate patch release of `bioformats2raw 0.5.1`, maybe alongside https://github.com/glencoesoftware/bioformats2raw/pull/167 and optionally https://github.com/glencoesoftware/bioformats2raw/pull/169.

PR #172, opened 2022-11-01
Add progress API

melissalinkert:
This should make it easier to get the conversion progress, for applications that use bioformats2raw as a library.

The test in e36f668 shows how a calling application is meant to use this.

`bioformats2raw -p ...` should continue to work as before.

If/once we're happy with this, raw2ometiff could get the same treatment.


src/main/java/com/glencoesoftware/bioformats2raw/IProgressListener.java
```diff
@@ -0,0 +1,41 @@
...
+public interface IProgressListener extends EventListener {
+
+  /**
+   * Indicates the beginning of processing a particular resolution.
+   *
+   * @param series the series index being processed
+   * @param resolution the resolution index being processed
+   * @param tileCount the total number of tiles in this resolution
+   */
+  void notifyResolution(int series, int resolution, int tileCount);
```
sbesson, line 21: My immediate feedback is that the contract of the notification events were not immediately clear just based on their names.  thought `notifyDone` would be a single event at the end of the conversion and I was confused by the fact `notifyResolution` indicates the start of the processing for a given resolution while `notifyChunk` indicates the end of the processing for a chunk.

A proposal to address this would be to name these methods e.g. as `notify[Series|Resolution|Chunk][Start|End]`. Not all these need to be defined initially but that would allow more granularity to be added in the future.

sbesson (approved):
Tested the command-line with the following test file

```
% cat test.fake.ini 
series=4
sizeX=30400
sizeY=25200
```

When calling the appropriate flag 
```
% ./gradlew run --args "test.fake.ini test.zarr -p"
```

The behavior is unchanged compared to `v0.6.0` with each resolution of each series having its own progress bar including the total number of resolutions to be converted.

When calling the default

```
% ./gradlew run --args "test.fake.ini test.zarr "
```

the output is being generated without any information passed to the terminal as previously.

Additionally, the new interface makes sense to me should allow other clients to inject their own implementation consuming different events. Happy for this to be included in the next minor version increment of `bioformats2raw`

sbesson (approved):

melissalinkert:
In preparation for 0.7.0, planning to merge this Monday (April 24) unless any further comments.

PR #173, opened 2022-11-01
Remove the "--pixel-type" option

### LINKED ISSUE #154: `--pixel-type` behaves incorrectly
The float to integer conversion introduced in #53 either needs a rewrite or to be removed.

Pixel type conversion is done as each tile is processed, which means that each tile (single channel for FL data) is handled separately:

https://github.com/glencoesoftware/bioformats2raw/blob/master/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java#L1008

The pixels are actually changed here:

https://github.com/glencoesoftware/bioformats2raw/blob/master/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java#L1805

What that does is take the original floating point tile and normalize the pixel values so that the minimum pixel value in the tile is 0.0 and the maximum pixel value is 1.0. Then each pixel value is multiplied by the output pixel type's maximum range, so that for uint16 you would get pixel values between 0 and 65535. That does a better job of preserving the pixel intensity range than just casting float values to a smaller type, but it means that different tiles will be normalized differently. This is more obvious when looking at multiple channels with different min/max ranges in the input image.

I don't think there is a way to fix this without knowing in advance what the overall min/max pixel values are for the entire input dataset.

/cc @mgheirat

melissalinkert:
Fixes #154.

As described in #154, `--pixel-type` never worked quite right, and we now have other ways to solve the problems it was meant to fix. This removes the option entirely (no deprecation), but could be changed to keep the option a no-op if anyone feels strongly.

Maybe worth including in the next patch release along with https://github.com/glencoesoftware/bioformats2raw/pull/171#pullrequestreview-1163027738?

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -36,7 +36,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.IntStream;
 
 import loci.common.Constants;
-import loci.common.DataTools;
 import loci.common.image.IImageScaler;
 import loci.common.image.SimpleImageScaler;
 import loci.common.services.DependencyException;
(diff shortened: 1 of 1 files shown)


sbesson:
> we now have other ways to solve the problems it was meant to fix.

Is this primarily referring to the min/max calculation per channel which is now happening by default in `bioformats2raw 0.5.0`? From a client perspective, this should definitely make it easier to rescale the data without having to modify the pixel type during conversion.

> This removes the option entirely (no deprecation), but could be changed to keep the option a no-op if anyone feels strongly.

Re-reading the original issue, I think we are talking about a data loss issue which is arguably one of the worst kind in a conversion library. I would also support dropping the option entirely.

erindiel (approved):

PR #177, opened 2022-11-21
Pyramid TIFF: ignore files where the first/last IFDs have different pixel types

melissalinkert:
This is meant to fix an edge case where there are multiple IFDs in descending size order that could be mistaken for a pyramid. The IFDs had a mix of uint8, uint8 RGB, and uint16 data.

Adding the `--extra-readers` option was a work-around in the meantime.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,6 @@ jobs:
       - name: Upload artifacts
         uses: actions/upload-artifact@v2
         with:
-          name: bioformats2raw
+          name: bioformats2raw ${{ matrix.java }}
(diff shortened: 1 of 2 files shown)


melissalinkert:
09580a8 is a bonus fix for GitHub Actions artifact uploading weirdness, since apparently that's what we're doing today. /cc @sbesson

erindiel (approved):
Internal sample data with this issue have multiple series including slide label, overview, and scan image. Without this PR, there is only a single series written to the Zarr output.

With these changes, tests across two sample datasets produce the expected number of series in the Zarr output, each with correct dimensions and pyramid resolutions.

Issue #168, opened 2022-10-18
Don't package logback.xml in bioformats2raw*.jar


melissalinkert:
See https://github.com/glencoesoftware/NGFF-Converter/pull/30#issuecomment-1282168128

PR #190, opened 2023-01-31
Add getters and setters for each option

melissalinkert:
As demonstrated by the new `testOptionsAPI` test, this should allow NGFF-Converter and other applications to configure the converter without constructing a list of command line arguments.

Prompted by https://github.com/glencoesoftware/NGFF-Converter/issues/39 and a discussion with @erindiel, @muhanadz, @sbesson, and @emilroz.

The diff looks like a lot, but really all this is doing is:
* adding a getter for each option's internal variable
* adding a setter for each option's internal variable, with some minimal input validation
* moving the picocli `@Option` and `@Parameter` annotations from the variables to the setters
* adding a test that uses the new setters/getters instead of command line arguments

I had hoped that there was a nicer way to connect picocli options to corresponding API methods, but as far as I can tell this is the least bad approach. `Converter` is getting to be quite large with these changes, so for a future PR it may be worth refactoring.

It's probably worth double-checking `bioformats2raw --help` and a few conversions with options you would normally use (or recommend to others). If there are additional unit tests that would be helpful, let me know and I can add them.

I'd expect this PR to conflict with #172, but this PR is higher priority.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -221,8 +221,9 @@ public class Converter implements Callable<Void> {
     arity = "1",
     description = "file to convert"
   )
-  public void setInputPath(Path input) {
-    inputPath = input;
+  public void setInputPath(String input) {
+    // this could be expanded later to support files not on disk
+    inputPath = Paths.get(input);
   }
 
   /**
@@ -786,8 +787,8 @@ public class Converter implements Callable<Void> {
   /**
    * @return path to input data
    */
-  public Path getInputPath() {
-    return inputPath;
+  public String getInputPath() {
+    return inputPath.toString();
   }
 
   /**
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -104,7 +104,7 @@ import ucar.ma2.InvalidRangeException;
 /**
  * Command line tool for converting whole slide imaging files to Zarr.
  */
-public class Converter implements Callable<Void> {
+public class Converter implements Callable<Integer> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(Converter.class);
 
@@ -1027,8 +1027,13 @@ public class Converter implements Callable<Void> {
 
   // Conversion methods
 
+  /**
+   * @return 0 if conversion completed without error,
+   *         -1 if conversion was not performed
+   * @throws Exception on most conversion errors
+   */
   @Override
-  public Void call() throws Exception {
+  public Integer call() throws Exception {
     if (printVersion) {
       String version = Optional.ofNullable(
         this.getClass().getPackage().getImplementationVersion()
@@ -1036,7 +1041,7 @@ public class Converter implements Callable<Void> {
       System.out.println("Version = " + version);
       System.out.println("Bio-Formats version = " + FormatTools.VERSION);
       System.out.println("NGFF specification version = " + NGFF_VERSION);
-      return null;
+      return -1;
     }
 
     if (inputPath == null) {
@@ -1109,7 +1114,7 @@ public class Converter implements Callable<Void> {
     executor = new ThreadPoolExecutor(
       maxWorkers, maxWorkers, 0L, TimeUnit.MILLISECONDS, queue);
     convert();
-    return null;
+    return 0;
   }
 
   /**
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -319,6 +319,9 @@ public class Converter implements Callable<Integer> {
     if (width > 0) {
       tileWidth = width;
     }
+    else {
+      LOGGER.warn("Ignoring invalid tile width: {}", width);
+    }
   }
 
   /**
@@ -338,6 +341,9 @@ public class Converter implements Callable<Integer> {
     if (height > 0) {
       tileHeight = height;
     }
+    else {
+      LOGGER.warn("Ignoring invalid tile height: {}", height);
+    }
   }
 
   /**
@@ -353,6 +359,9 @@ public class Converter implements Callable<Integer> {
   public void setChunkDepth(int depth) {
     if (depth > 0) {
       chunkDepth = depth;
+    }
+    else {
+      LOGGER.warn("Ignoring invalid chunk depth: {}", depth);
     }
   }
 
@@ -419,6 +428,9 @@ public class Converter implements Callable<Integer> {
   public void setMaxWorkers(int workers) {
     if (workers > 0) {
       maxWorkers = workers;
+    }
+    else {
+      LOGGER.warn("Ignoring invalid worker count: {}", workers);
     }
   }
 
@@ -654,6 +666,9 @@ public class Converter implements Callable<Integer> {
     if (tileFill == null || (tileFill >= 0 && tileFill <= 255)) {
       fillValue = tileFill;
     }
+    else {
+      LOGGER.warn("Ignoring invalid fill value (must be 0-255): {}", tileFill);
+    }
   }
 
   /**
@@ -761,6 +776,9 @@ public class Converter implements Callable<Integer> {
   public void setMinImageSize(int min) {
     if (min > 0) {
       minSize = min;
+    }
+    else {
+      LOGGER.warn("Ignoring invalid minimum image size: {}", min);
     }
   }
 
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -128,282 +128,54 @@ public class Converter implements Callable<Void> {
   /** NGFF specification version.*/
   public static final String NGFF_VERSION = "0.4";
 
-  @Parameters(
-    index = "0",
-    arity = "1",
-    description = "file to convert"
-  )
   private volatile Path inputPath;
-
-  @Parameters(
-    index = "1",
-    arity = "1",
-    description = "path to the output pyramid directory. " +
-      "The given path can also be a URI (containing ://) " +
-      "which will activate **experimental** support for " +
-      "Filesystems. For example, if the output path given " +
-      "is 's3://my-bucket/some-path' *and* you have an "+
-      "S3FileSystem implementation in your classpath, then " +
-      "all files will be written to S3."
-  )
   private volatile String outputLocation;
 
-  @Option(
-    names = "--output-options",
-    split = "\\|",
-    description = "|-separated list of key-value pairs " +
-      "to be used as an additional argument to Filesystem " +
-      "implementations if used. For example, " +
-      "--output-options=s3fs_path_style_access=true|... " +
-      "might be useful for connecting to minio."
-  )
   private Map<String, String> outputOptions;
-
-  @Option(
-    names = {"-r", "--resolutions"},
-    description = "Number of pyramid resolutions to generate"
-  )
   private volatile Integer pyramidResolutions;
-
-  @Option(
-    names = {"-s", "--series"},
-    arity = "0..1",
-    split = ",",
-    description = "Comma-separated list of series indexes to convert"
-  )
   private volatile List<Integer> seriesList = new ArrayList<Integer>();
 
-  @Option(
-    names = {"-w", "--tile_width"},
-    description = "Maximum tile width (default: ${DEFAULT-VALUE}). " +
-      "This is both the chunk size (in X) when writing Zarr and the tile " +
-      "size used for reading from the original data. Changing the tile " +
-      "size may have performance implications."
-  )
   private volatile int tileWidth = 1024;
-
-  @Option(
-    names = {"-h", "--tile_height"},
-    description = "Maximum tile height (default: ${DEFAULT-VALUE}). " +
-      "This is both the chunk size (in Y) when writing Zarr and the tile " +
-      "size used for reading from the original data. Changing the tile " +
-      "size may have performance implications."
-  )
   private volatile int tileHeight = 1024;
-
-  @Option(
-    names = {"-z", "--chunk_depth"},
-    description = "Maximum chunk depth to read (default: ${DEFAULT-VALUE}) "
-  )
   private volatile int chunkDepth = 1;
-
-  @Option(
-    names = {"--log-level", "--debug"},
-    arity = "0..1",
-    description = "Change logging level; valid values are " +
-      "OFF, ERROR, WARN, INFO, DEBUG, TRACE and ALL. " +
-      "(default: ${DEFAULT-VALUE})",
-    fallbackValue = "DEBUG"
-  )
   private volatile String logLevel = "WARN";
-
-  @Option(
-    names = {"-p", "--progress"},
-    description = "Print progress bars during conversion",
-    help = true
-  )
   private volatile boolean progressBars = false;
-
-  @Option(
-    names = "--version",
-    description = "Print version information and exit",
-    help = true
-  )
   private volatile boolean printVersion = false;
 
-  @Option(
-    names = "--max_workers",
-    description = "Maximum number of workers (default: ${DEFAULT-VALUE})"
-  )
   // cap the default worker count at 4, to prevent problems with
   // large images that are not tiled
   private volatile int maxWorkers =
       (int) Math.min(4, Runtime.getRuntime().availableProcessors());
 
-  @Option(
-    names = "--max_cached_tiles",
-    description =
-      "Maximum number of tiles that will be cached across all "
-      + "workers (default: ${DEFAULT-VALUE})"
-  )
   private volatile int maxCachedTiles = 64;
-
-  @Option(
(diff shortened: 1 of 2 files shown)


sbesson (changes requested):
Generally, the spirit of this PR goes alongside recent improvements in #168 as the usage of this library goes beyond the initial scope of a standalone command-line converter and evolves towards an API that can be used by downstream component like NGFF-Converter.

I had separately briefly investigated the picocli built-in API to introspect and manipulate options but it's either limited or requires to use the internal model which sounds very brittle. In that sense, exposing a proper API allowing to set and retrieve conversion parameters and chain steps together sounds like the correct approach.

As noted in the description, the primary downside of this PR is the increase in terms of lines of code especially as every new property requires a private field, a getter and a setter. There are also different schools of thoughts on the different strateg an annotating public fields would definitely be the change with the least amount of code.
One advantage of the getter/setter approach quickly mentioned in the description is the ability to add validation logic. Should logging warning be emitted when some input does not match the setter condition and is ignored?

In terms of functional testing, I tried to combine this together with https://github.com/glencoesoftware/raw2ometiff/pull/97 and consume then within NGFF-Converter which is the primary driver for this features, more specifically https://github.com/glencoesoftware/NGFF-Converter/issues/39. After building and publishing both libraries, NGFF-Converter was built with the following diff

```diff
diff --git a/build.gradle b/build.gradle
index 5b3b5a0..15b6fff 100644
--- a/build.gradle
+++ b/build.gradle
@@ -6,8 +6,8 @@ plugins {
 
 version = '1.1.4'
 String bfversion = "6.12.0"
-String b2rversion = "0.6.1"
-String r2oversion = "0.4.1"
+String b2rversion = "0.7.0-SNAPSHOT"
+String r2oversion = "0.5.0-SNAPSHOT"
 
 mainClassName = 'com.glencoesoftware.convert.Launcher'
 applicationName = 'NGFF-Converter'
diff --git a/src/main/java/com/glencoesoftware/convert/ConverterTask.java b/src/main/java/com/glencoesoftware/convert/ConverterTask.java
index 9608df7..2de9bdd 100644
--- a/src/main/java/com/glencoesoftware/convert/ConverterTask.java
+++ b/src/main/java/com/glencoesoftware/convert/ConverterTask.java
@@ -9,6 +9,7 @@ package com.glencoesoftware.convert;
 
 import ch.qos.logback.classic.Level;
 import com.glencoesoftware.bioformats2raw.Converter;
+import com.glencoesoftware.bioformats2raw.ZarrCompression;
 import com.glencoesoftware.pyramid.PyramidFromDirectoryWriter;
 import javafx.application.Platform;
 import javafx.concurrent.Task;
@@ -49,9 +50,7 @@ class ConverterTask extends Task<Integer> {
         RunnerExecutionExceptionHandler runHandler = new RunnerExecutionExceptionHandler();
         int count = 0;
         for (IOPackage job : inputFileList.getItems()) {
-            CommandLine runner = new CommandLine(new Converter());
-            runner.setParameterExceptionHandler(paramHandler);
-            runner.setExecutionExceptionHandler(runHandler);
+            Converter ngffConverter = new Converter();
             File in = job.fileIn;
             File out;
             if (this.interrupted || job.status == COMPLETED || job.status == ERROR) {
@@ -75,20 +74,26 @@ class ConverterTask extends Task<Integer> {
                     temporaryStorage = job.fileOut.getParent();
                 }
                 out = new File(Paths.get(temporaryStorage, UUID.randomUUID() + ".zarr").toString());
-                Set<String> validArgs = runner.getCommandSpec().optionsMap().keySet();
-                List<String> argsToUse = args.stream().filter(
-                        (arg) -> validArgs.contains(arg.split("=")[0])).toList();
-                params = new ArrayList<>(argsToUse);
+                // Set<String> validArgs = runner.getCommandSpec().optionsMap().keySet();
+                // List<String> argsToUse = args.stream().filter(
+                //         (arg) -> validArgs.contains(arg.split("=")[0])).toList();
+                // params = new ArrayList<>(argsToUse);
             } else {
                 out = job.fileOut;
-                params = new ArrayList<>(args);
+                // params = new ArrayList<>(args);
             }
-
             // Construct args list
+            params = new ArrayList<>(args);
             params.add(0, out.getAbsolutePath());
             params.add(0, in.getAbsolutePath());
             String[] fullParams = params.toArray(new String[args.size()]);
-
+            
+            ngffConverter.setInputPath(Paths.get(in.getAbsolutePath()));
+            ngffConverter.setOutputPath(out.getAbsolutePath());
+            ngffConverter.setCompression(ZarrCompression.raw);
+            CommandLine ngffRunner = new CommandLine(ngffConverter);
+            ngffRunner.setParameterExceptionHandler(paramHandler);
+            ngffRunner.setExecutionExceptionHandler(runHandler);
 
             int result;
             if (in.getAbsolutePath().endsWith(".zarr")) {
@@ -96,24 +101,28 @@ class ConverterTask extends Task<Integer> {
                 result = 0;
                 out = job.fileIn;
             } else {
-                LOGGER.info("Executing bioformats2raw with args " + Arrays.toString(fullParams));
-                result = runner.execute(fullParams);
+                // LOGGER.info("Executing bioformats2raw with args " + Arrays.toString(fullParams));
+                result = ngffRunner.execute(fullParams);
             }

              if (result == 0 && job.outputMode == PrimaryController.OutputMode.TIFF)  {
                 LOGGER.info("NGFF intermediate generated, converting to TIFF");
-                CommandLine converter = new CommandLine(new PyramidFromDirectoryWriter());
-                converter.setParameterExceptionHandler(paramHandler);
-                converter.setExecutionExceptionHandler(runHandler);
-                Set<String> validConverterArgs = converter.getCommandSpec().optionsMap().keySet();
-                List<String> convertArgs = args.stream().filter(
-                        (arg) -> validConverterArgs.contains(arg.split("=")[0])).toList();
-                ArrayList<String> convertParams = new ArrayList<>(convertArgs);
+                PyramidFromDirectoryWriter ometiffConverter = new PyramidFromDirectoryWriter();
+                ometiffConverter.setInputPath(Paths.get(out.getAbsolutePath()));
+                ometiffConverter.setOutputPath(Paths.get(job.fileOut.getAbsolutePath()));
+                ometiffConverter.setCompression("jpeg");
+                // Set<String> validConverterArgs = converter.getCommandSpec().optionsMap().keySet();
+                // List<String> convertArgs = args.stream().filter(
+                //         (arg) -> validConverterArgs.contains(arg.split("=")[0])).toList();
+                ArrayList<String> convertParams = new ArrayList<>();
                 convertParams.add(0, job.fileOut.getAbsolutePath());
                 convertParams.add(0, out.getAbsolutePath());
                 String[] phase2Params = convertParams.toArray(new String[0]);
-                LOGGER.info("Executing raw2ometiff with args " + Arrays.toString(phase2Params));
-                result = converter.execute(phase2Params);
+                // LOGGER.info("Executing raw2ometiff with args " + Arrays.toString(phase2Params));
+                CommandLine ometiffRunner = new CommandLine(ometiffConverter);
+                ometiffRunner.setParameterExceptionHandler(paramHandler);
+                ometiffRunner.setExecutionExceptionHandler(runHandler);
+                result = ometiffRunner.execute(phase2Params);
                 LOGGER.info("Cleaning up intermediate files");
                 if (out != job.fileIn) {
                     FileUtils.deleteDirectory(out);
```

The tests demonstrate the usage of the `Converter` callable directly using the setter/getter. Howeer these setters do not seem to be recognized when wrapping the class within a `CommandLine` API:
-  I received a `picocli.CommandLine$MissingParameterException: Missing required parameters: '<inputPath>', '<outputPath>'` and needed to pass the input/output as a string array like previously
- the compression argument were also ignored
Does that mean `CommandLine` should exclusively be used with a command-line utility usage in mind and the converter should use the callable directly?

Additionally exposing the setters as APIs reveals some inconsistencies across the converters. Listing a few of these inconsistencies below and we can decide for each of them to address them immediately, discard them and/or capture them as issues:
- `setInputPath/setOutputPath` use `Path` as input except for `Converter.setOutputPath` which is a `String`
- `setCompression` takes an enum as input for `Converter` but a string for `PyramidFromDirectoryWriter`
- `PyramidFromDirectoryWriter` uses `setCompressionQuality` which could be likely be extended to support additional options like `Converter.setCompressionProperties`

sbesson (changes requested):
Retested with the latest changes in this PR as well as https://github.com/glencoesoftware/raw2ometiff/pull/97 and 

```diff
diff --git a/build.gradle b/build.gradle
index 5b3b5a0..15b6fff 100644
--- a/build.gradle
+++ b/build.gradle
@@ -6,8 +6,8 @@ plugins {
 
 version = '1.1.4'
 String bfversion = "6.12.0"
-String b2rversion = "0.6.1"
-String r2oversion = "0.4.1"
+String b2rversion = "0.7.0-SNAPSHOT"
+String r2oversion = "0.5.0-SNAPSHOT"
 
 mainClassName = 'com.glencoesoftware.convert.Launcher'
 applicationName = 'NGFF-Converter'
diff --git a/src/main/java/com/glencoesoftware/convert/ConverterTask.java b/src/main/java/com/glencoesoftware/convert/ConverterTask.java
index 9608df7..7600ff0 100644
--- a/src/main/java/com/glencoesoftware/convert/ConverterTask.java
+++ b/src/main/java/com/glencoesoftware/convert/ConverterTask.java
@@ -9,6 +9,7 @@ package com.glencoesoftware.convert;
 
 import ch.qos.logback.classic.Level;
 import com.glencoesoftware.bioformats2raw.Converter;
+import com.glencoesoftware.bioformats2raw.ZarrCompression;
 import com.glencoesoftware.pyramid.PyramidFromDirectoryWriter;
 import javafx.application.Platform;
 import javafx.concurrent.Task;
@@ -44,14 +45,10 @@ class ConverterTask extends Task<Integer> {
     }
 
     @Override
-    protected Integer call() throws IOException {
-        RunnerParameterExceptionHandler paramHandler = new RunnerParameterExceptionHandler();
-        RunnerExecutionExceptionHandler runHandler = new RunnerExecutionExceptionHandler();
+    protected Integer call() throws IOException, Exception {
         int count = 0;
         for (IOPackage job : inputFileList.getItems()) {
-            CommandLine runner = new CommandLine(new Converter());
-            runner.setParameterExceptionHandler(paramHandler);
-            runner.setExecutionExceptionHandler(runHandler);
+            Converter ngffConverter = new Converter();
             File in = job.fileIn;
             File out;
             if (this.interrupted || job.status == COMPLETED || job.status == ERROR) {
@@ -75,20 +72,13 @@ class ConverterTask extends Task<Integer> {
                     temporaryStorage = job.fileOut.getParent();
                 }
                 out = new File(Paths.get(temporaryStorage, UUID.randomUUID() + ".zarr").toString());
-                Set<String> validArgs = runner.getCommandSpec().optionsMap().keySet();
-                List<String> argsToUse = args.stream().filter(
-                        (arg) -> validArgs.contains(arg.split("=")[0])).toList();
-                params = new ArrayList<>(argsToUse);
             } else {
                 out = job.fileOut;
-                params = new ArrayList<>(args);
             }
 
-            // Construct args list
-            params.add(0, out.getAbsolutePath());
-            params.add(0, in.getAbsolutePath());
-            String[] fullParams = params.toArray(new String[args.size()]);
-
+            ngffConverter.setInputPath(in.getAbsolutePath());
+            ngffConverter.setOutputPath(out.getAbsolutePath());
+            ngffConverter.setCompression(ZarrCompression.raw);
 
             int result;
             if (in.getAbsolutePath().endsWith(".zarr")) {
@@ -96,24 +86,18 @@ class ConverterTask extends Task<Integer> {
                 result = 0;
                 out = job.fileIn;
             } else {
-                LOGGER.info("Executing bioformats2raw with args " + Arrays.toString(fullParams));
-                result = runner.execute(fullParams);
+                ngffConverter.call();
+                result = 0;
             }
 
             if (result == 0 && job.outputMode == PrimaryController.OutputMode.TIFF)  {
                 LOGGER.info("NGFF intermediate generated, converting to TIFF");
-                CommandLine converter = new CommandLine(new PyramidFromDirectoryWriter());
-                converter.setParameterExceptionHandler(paramHandler);
-                converter.setExecutionExceptionHandler(runHandler);
-                Set<String> validConverterArgs = converter.getCommandSpec().optionsMap().keySet();
-                List<String> convertArgs = args.stream().filter(
-                        (arg) -> validConverterArgs.contains(arg.split("=")[0])).toList();
-                ArrayList<String> convertParams = new ArrayList<>(convertArgs);
-                convertParams.add(0, job.fileOut.getAbsolutePath());
-                convertParams.add(0, out.getAbsolutePath());
-                String[] phase2Params = convertParams.toArray(new String[0]);
-                LOGGER.info("Executing raw2ometiff with args " + Arrays.toString(phase2Params));
-                result = converter.execute(phase2Params);
+                PyramidFromDirectoryWriter ometiffConverter = new PyramidFromDirectoryWriter();
+                ometiffConverter.setInputPath(out.getAbsolutePath());
+                ometiffConverter.setOutputPath(job.fileOut.getAbsolutePath());
+                ometiffConverter.setCompression("JPEG");
+                ometiffConverter.call();
+                result = 0;
                 LOGGER.info("Cleaning up intermediate files");
                 if (out != job.fileIn) {
                     FileUtils.deleteDirectory(out);
```

With this sequence of calls, I was able to convert a fake file to OME-NGFF and/or OME-TIFF with various compression schemes (uncompresed for Zarr, zlib or JPEG for TIFF).
From the diff above, the outstanding requirement is that in the absence of the `CommandLine` wrapper, we need some mechanism to communicate the conversion status. Immediate suggestion would be to update the converter class to implement `Callable<Integer>` and have `call` return the exit code.

melissalinkert:
I'd expect `call()` to throw an exception on pretty much any failed conversion, but a8ac114 now returns an exit code. Right now that is 0 for a successful conversion and -1 if version information was requested.

sbesson:
I am having second thoughts on the  value of the `Callable<Integer>` change. My request was largely driven by the current implementation of NGFF-Converter which consumes the return value from the `CommandLine` class - see https://picocli.info/quick-guide.html#_exception_exit_codes.

Reading the documentation, a `CommandLine` can wrap any `Callable` or `Runnable` which do not return a result. Using exceptions to communicate failed processes is inline with the expectation of the command-line wrapper. This leaves the question of whether we would like to differentiate the output of `--version` from any other conversion. How do you feel about it ?

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -128,13 +128,107 @@ public class Converter implements Callable<Void> {
   /** NGFF specification version.*/
   public static final String NGFF_VERSION = "0.4";
 
+  private volatile Path inputPath;
+  private volatile String outputLocation;
```
sbesson, line 132: Any reason to keep `outputLocation` or is it also worth unifying the internal variable name with the getter/setter?

melissalinkert: `outputPath` is defined later on, and is resolved from `outputLocation` as either a path on disk or an s3 location. I can rename `outputLocation` to `outputPath` if you feel that's clearer, but then would need to rename the current `outputPath` to something else.

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -157,139 +258,297 @@ public class Converter implements Callable<Void> {
...
       "size used for reading from the original data. Changing the tile " +
-      "size may have performance implications."
+      "size may have performance implications.",
+    defaultValue = "1024"
   )
-  private volatile int tileWidth = 1024;
+  public void setTileWidth(int width) {
+    if (width > 0) {
+      tileWidth = width;
+    }
```
sbesson, line 321: Could we add a warning in case the value is ignored (similar to resolution above)?

src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -157,139 +258,297 @@ public class Converter implements Callable<Void> {
...
     names = {"-z", "--chunk_depth"},
-    description = "Maximum chunk depth to read (default: ${DEFAULT-VALUE}) "
+    description = "Maximum chunk depth to read (default: ${DEFAULT-VALUE}) ",
+    defaultValue = "1"
   )
-  private volatile int chunkDepth = 1;
+  public void setChunkDepth(int depth) {
+    if (depth > 0) {
+      chunkDepth = depth;
+    }
```
sbesson, line 356: Same as above

sbesson (changes requested):
Successfully retested the following parameters via command-line
- resolutions
- series
- tile width/height 
- log-level
- progress bars
- compression

A couple of last suggestions to unify the logging behavior when invalid values are ignored in the setters.

melissalinkert:
2234a9e should address all of the logging comments.

DavidStirling:
Looks reasonable to me, is there any way to fetch the CLI parameter descriptions out when using API rather than CLI mode? Having access to those would save having to duplicate the text when putting together a GUI.

melissalinkert:
> Looks reasonable to me, is there any way to fetch the CLI parameter descriptions out when using API rather than CLI mode?

This PR doesn't attempt to do anything on that front, but it's a good idea for a next step. Moving the descriptions to a resource bundle as described here:

https://picocli.info/#_description_text_in_resourcebundle
https://picocli.info/#_internationalization

might be one way to do this.

PR #191, opened 2023-02-10
Add a MapAnnotation describing the conversion

melissalinkert:
Discussed with @erindiel, @emilroz, and @sbesson earlier today that it might be useful to have a record of the input file name and converter version(s). Not expecting this to be merged as-is, but it's a place to continue that discussion.

With this PR, `bioformats2raw test.fake test.zarr` should include the following annotation in `test.zarr/OME/METADATA.ome.xml`:

```
<MapAnnotation ID="Annotation:bioformats2raw:0" Namespace="glencoesoftware.com/ngff/provenance">
  <Value>
    <M K="input_path">test.fake</M>
    <M K="bioformats2raw">0.7.0-SNAPSHOT</M>
    <M K="bioformats">6.11.1</M>
  </Value>
</MapAnnotation>
```

Definitely happy to hear other ideas or better naming suggestions. Maybe the Bio-Formats version doesn't need to be stated explicitly since it can be inferred from the bioformats2raw version?

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -695,6 +695,12 @@ public class Converter implements Callable<Void> {
           meta.setMapAnnotationNamespace(PROVENANCE_NAMESPACE, annotationIndex);
           meta.setMapAnnotationValue(annotation, annotationIndex);
 
+          if (meta.getCreator() != null) {
+            LOGGER.warn("Overwriting OME-XML Creator attribute: {}",
+              meta.getCreator());
+          }
+          meta.setCreator("bioformats2raw " + getVersion());
+
           String xml = service.getOMEXML(meta);
 
           // write the original OME-XML to a file
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -1296,8 +1296,8 @@ public class Converter implements Callable<Integer> {
 
           List<MapPair> annotation = new ArrayList<MapPair>();
           annotation.add(new MapPair("input_path", inputPath.toString()));
-          annotation.add(new MapPair("bioformats2raw", getVersion()));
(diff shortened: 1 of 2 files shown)


joshmoore:
Just as points of comparison, the list of fields from omero-cli-transfer for this kind of provenance map annotation are:  `timestamp`, `software`, `version`, and `md5` along with the OMERO specific `hostname`, `db_id`, `orig_user`, `orig_group`, and `img_id`.

see: https://github.com/ome/omero-cli-transfer/blob/main/src/generate_omero_objects.py#L66

and just as a side note: one could conceivably end up with multiple of these through a PFF->Zarr-TIFF->Zarr loop.

sbesson:
Adding to list of options, [OME.Creator](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#OME_Creator) is used by the Bio-Formats command line tools and could also be another place to store this information. It is a simple string and thus has natural limitations to describe multiple conversion but possibly the last converter could be stored there in a serialized form (e.g. `bioformats2raw x.y.z`).

> and just as a side note: one could conceivably end up with multiple of these through a PFF->Zarr-TIFF->Zarr loop.

:+1: to being able to handle multiple provenance annotations. Within the scope of Glencoe utilities, I assume a bioformats2raw+raw2ometiff pipeline should generate an OME-TIFF that contains metadata about both steps.

sbesson (commented):
A few thoughts:
- I find `bioformats` and` bioformats2raw` are redundant as the former can be inferred from the version of the latter
- looking at other tools trying to capture similar provenance metadata like `omero-cli-transfer` discussed above or RO-Crate which uses the schema.org [SoftwareApplication](https://schema.org/SoftwareApplication) type, name and version will typically be represented as separate key/value pairs
- thinking of how to chain several instances of these provenance annotations, a couple of options:
  - a `ListAnnotation` with reference to each of the individual `MapAnnotation` describing the software
  - several `MapAnnotation` with different namespaces 
  - a single `MapAnnotation` with ordered key/value pairs

PR #203, opened 2023-05-19
Update to new jzarr 0.4.2

melissalinkert:
Prompted by https://github.com/ome/ZarrReader/pull/54, see https://github.com/zarr-developers/jzarr.

Discussed briefly with @sbesson. I am not suggesting that we consider this for bioformats2raw 0.7.0, it's more a place to start. Whatever we decide to do here should likely be applied to other repos where we use jzarr (including but not limited to raw2ometiff).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,16 +22,10 @@ repositories {
         url 'https://repo.glencoesoftware.com/repository/n5-zarr-snapshots/'
     }
     maven {
-        url 'https://repo.glencoesoftware.com/repository/jzarr-snapshots'
-    }
-    maven {
         url 'https://artifacts.glencoesoftware.com/artifactory/ome.external'
     }
     maven {
         url 'https://artifacts.glencoesoftware.com/artifactory/scijava-thirdparty'
-    }
-    maven {
-        url 'https://artifacts.glencoesoftware.com/artifactory/jzarr-releases'
     }
     maven {
         name 'Unidata'
--- a/build.gradle
+++ b/build.gradle
@@ -49,7 +49,7 @@ dependencies {
     implementation 'ome:formats-gpl:6.13.0'
     implementation 'info.picocli:picocli:4.6.1'
     implementation 'com.univocity:univocity-parsers:2.8.4'
-    implementation 'com.bc.zarr:jzarr:0.3.5'
+    implementation 'dev.zarr:jzarr:0.3.7'
     // implementation 'org.carlspring.cloud.aws:s3fs-nio:1.0-SNAPSHOT'
     // implementation 'io.nextflow:nxf-s3fs:1.1.0'
     implementation 'org.lasersonlab:s3fs:2.2.3'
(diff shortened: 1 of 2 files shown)


sbesson (commented):
In discussing with @melissalinkert about these changes, noting do

- the `com.bc.zarr:jzarr` library should be considered as inactive
- the cleanup of all intermediate repositories and the ability to use Maven Central is really nice
- looking nature of the changes between the 0.3.5 and the 0.4.0 changes https://github.com/bcdev/jzarr/compare/0.3.5...zarr-developers:jzarr:0.4.0, I see nothing of particular concerns:
   - essential metadata got added to the POM like the license :+1:
   - a few API additions which could be consumed by the converter e.g. the multi-threaded blosc compression
   - the bump of the jackson dependency which is overriden by a more recent versions from `ome:formats-gpl` anyways
- at present, https://github.com/zarr-developers/jzarr feels like the best place to discuss & propose new API additions which would be relevant to the needs of this library
- we will likely need to unify the change across all conversion utilities to avoid a mixture of `jzarr` 

Could we update
https://github.com/glencoesoftware/bioformats2raw/blob/b7c5440fc79debbcd9bff965377e78f61f963b9a/src/test/java/com/glencoesoftware/bioformats2raw/test/ZarrTest.java#L270-L273 to use the new dimension separator getter available in the new version?

sbesson:
:+1: maybe let's just remove the outdated test comment and then this should be ready
```
   // no getter for DimensionSeparator in ZarrArray
    // check that the correct separator was used by checking
    // that the expected first chunk file exists
```

Next question is whether we want to plan a 0.7.0 release or a release candidate that can be consumed in all the related converter utility PRs adjusting the dependency to use `dev.zarr`

melissalinkert:
Test comment updated in 609f092. As discussed separately, this PR *should not* be merged right now, pending investigation of https://github.com/zarr-developers/jzarr/issues/14.

build.gradle
```diff
@@ -49,7 +40,7 @@ dependencies {
     implementation 'ome:formats-gpl:6.13.0'
     implementation 'info.picocli:picocli:4.6.1'
     implementation 'com.univocity:univocity-parsers:2.8.4'
-    implementation 'com.bc.zarr:jzarr:0.3.5'
+    implementation 'dev.zarr:jzarr:0.4.0'
```
sbesson, line 43:
```suggestion
    implementation 'dev.zarr:jzarr:0.4.2'
```

joshmoore: :+1:

sbesson (approved):
Retested the last commit and confirmed that:
- the array metadata of the generated Zarr datasets no longer include the additional `numThreads` key
- the Zarr datasets can be opened both using the `zarr-python` implementation and another Java implementation e.g. `raw2ometiff`

A few recent maintenance issues (https://github.com/zarr-developers/jzarr/pull/17) are further reinforcing my feeling expressed in https://github.com/glencoesoftware/bioformats2raw/pull/203#pullrequestreview-1639353388 i.e. the inactive state of `com.bc.zarr:jzarr` effectively elevates the `dev.zarr:jzarr` fork as the new reference Java implementation for the Zarr v2 specification.

PR #213, opened 2023-08-11
Add support for INI files representing a Metaxpress plate

melissalinkert:
Be sure to check the `Channel.EmissionWavelength`, `Pixels.PhysicalSizeX`, `Pixels.PhysicalSizeY`, and `Plane.ExposureTime` values in the generated OME-XML. The logic there is slightly different, as it uses an `OMEXMLMetadata` to transport values instead of getters on `MetamorphReader`, as getters for those four values are not present in current Bio-Formats.


mabruce (approved):
`Channel.EmissionWavelength`, `Pixels.PhysicalSizeX`, `Pixels.PhysicalSizeY`, and `Plane.ExposureTime` are all present in the OME-XML. Conversion of a plate from the INI format to OME-NGFF works as expected.

emilroz (approved):

Issue #180, opened 2022-12-06
Convert to yx or cyx


LucaMarconato:
I am converting a yx ome.tif file to NGFF. Is there an option to create a yx or cyx image? I don't need the t and z dimensions.

melissalinkert:
@LucaMarconato : No, there is not an option to remove unused dimensions. bioformats2raw intentionally does not implement all features of the OME-NGFF specification, and assumes that all 5 dimensions should be preserved.

We're unlikely to implement this feature any time soon, but would be happy to look at a pull request if you (or anyone else) are interested in working on it.

LucaMarconato:
Thank you for the information!

jluethi:
Just adding my voice to that of @LucaMarconato & @d-v-b [over here](https://github.com/glencoesoftware/bioformats2raw/issues/209) to mention that it would be very useful to avoid adding such extra dimensions. An example of the issue this raises with other tools can be found in this discussion here: https://forum.image.sc/t/fractal-framework-zarr-compatibility/92536/10?u=jluethi.

It's not our top priority, so can't provide a PR for this anytime soon. But wanted to give a +1 for its usefulness if this issue gets raised in the future.

PR #215, opened 2023-08-29
RGB support

melissalinkert:
Adds a `--keep-rgb` option, which omits `ChannelSeparator` from the reader stack and will return RGB planes if that's what is stored in the input data. The reader's `isInterleaved()` is used in this case to determine the axis order; `isInterleaved() == true` means that the order will by `*YXC`. We may want to think about either forcing interleaved across the board if `--keep-rgb` is used, or adding a sub-option to do that.

The case where there are multiple channels each consisting of an RGB plane is not supported with `--keep-rgb`. Same for downsampling types that require OpenCV. We could probably add both, just keeping it simpler for now.

`DimensionOrder` in `METADATA.ome.xml` is not altered, so that comes directly from the underlying reader. With `--keep-rgb` and input RGB data, I'd expect `SamplesPerPixel` on each `Channel` to be greater than 1.

This is barely tested and has no unit tests yet. Just opening this as a draft now in case I run out of time to do more this week.

Discussed earlier today with @sbesson, cc @kkoz, @muhanadz

#180, https://github.com/glencoesoftware/isyntax2raw/pull/29, and https://github.com/glencoesoftware/raw2ometiff/issues/26 may be related.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -1656,10 +1656,30 @@ public class Converter implements Callable<Integer> {
         String.format(scaleFormatString,
             getScaleFormatStringArgs(series, resolution - 1));
     final ZarrArray zarr = ZarrArray.open(getRootPath().resolve(pathName));
+
     int[] dimensions = zarr.getShape();
     int[] blockSizes = zarr.getChunks();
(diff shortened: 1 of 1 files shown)


perlman:
I think we (with my JAX hat) would find this useful! We have a lot of RGB image data, and having an easy mechanism for the channels to be stored in the same chunk would certainly speed up visualization.

PR #216, opened 2023-09-13
Add a separate README section describing each of the additional readers

melissalinkert:
Fixes #214.

PR #218, opened 2023-10-02
Add Fluidigm Hyperion .mcd reader

melissalinkert:


sbesson:
NB: this might be another good candidate to add to the new section proposed in https://github.com/glencoesoftware/bioformats2raw/pull/216

emilroz:
Reader works well. One suggestion from me would be to move X,Y,Z channels to the end of the channel list.

melissalinkert:
> One suggestion from me would be to move X,Y,Z channels to the end of the channel list.

c6f46dc should do this; make sure to compare channel names and the actual images against ground truth referenced in the original issue.

emilroz (approved):

PR #226, opened 2023-11-15
Update setters to accept default values

melissalinkert:
...and define default values in the way that picocli expects them. Replaces #219, see discussion starting at https://github.com/glencoesoftware/bioformats2raw/pull/219#issuecomment-1773208406.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,7 +47,7 @@ configurations.all {
 dependencies {
     implementation 'net.java.dev.jna:jna:5.10.0'
     implementation 'ome:formats-gpl:7.0.0'
-    implementation 'info.picocli:picocli:4.6.1'
+    implementation 'info.picocli:picocli:4.7.5'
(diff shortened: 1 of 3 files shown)


DavidStirling (changes requested):
These changes do make the setters accept null values where needed, however because the setting values are no longer initialised by the `Converter` class there are no longer effective defaults outside of the CLI interface.

As a minimal example, I would expect that creating a Converter and setting input & output would be enough to run `.call()` with some standard settings assumed, however with this PR that method will now fail because a bunch of settings are `null` and this is unhandled within the executor.

I would suggest that either we continue initialising these settings on creation, or the Converter should accept `null` arguments when called and take that as a prompt to fetch the picocli default.

sbesson (changes requested):
This now requires conflicts to be merged as the version of Bio-Formats was bumped to 7.0.1. Then we should be in a good position to get this merged for `0.8.0`

sbesson (approved):

PR #230, opened 2023-12-01
Add getter for ThreadPoolExecutor

melissalinkert:
As discussed yesterday with @DavidStirling and @chris-allan (and see previous comments in glencoesoftware/raw2ometiff#116). This should allow more visibility into running threads that are performing tile conversions. `ExecutorService` alone doesn't provide most of the features that are likely to be useful, so `executor` is now explicitly a `ThreadPoolExecutor` so that we don't have to cast it elsewhere.


sbesson (approved):

DavidStirling (approved):

melissalinkert:
Closing as discussed just now with @DavidStirling, since this isn't especially helpful. We'll plan a more organized effort around this topic for the next release.

PR #83, opened 2021-02-27
Resolution-based progress bars

joshmoore:
Uses http://tongfei.me/progressbar/ to print progress
bars for each resolution to standard out. The default
logging level is set to WARN for optimal display when
there are no issues.

```
$ build/install/bioformats2raw/bin/bioformats2raw build/install/a.fake build/out
resolution 0 100% │█████████████████████████████████████████████████████████████████████████│ 2400/2400 (0:00:05 / 0:00:00)
resolution 1 100% │███████████████████████████████████████████████████████████████████████████│ 600/600 (0:00:03 / 0:00:00)
resolution 2 100% │███████████████████████████████████████████████████████████████████████████│ 150/150 (0:00:00 / 0:00:00)
resolution 3 100% │███████████████████████████████████████████████████████████████████████████│ 150/150 (0:00:00 / 0:00:00)
resolution 4 100% │███████████████████████████████████████████████████████████████████████████│ 150/150 (0:00:00 / 0:00:00)
```

PR #234, opened 2024-01-15
Fix progress and help options

melissalinkert:
As discussed earlier today with @sbesson and @chris-allan.

With this change, commands similar to:

```
$ bioformats2raw --help
$ bioformats2raw -p --foo test.fake test.zarr
$ bioformats2raw -p --help
```

should print a usage message and exit.

I don't know why the `help` flag was set for `-p`/`--progress-bars`, maybe @joshmoore remembers from the initial implementation in #83?

raw2ometiff will need a similar update once we're happy with the state of this PR.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -401,7 +401,6 @@ public class Converter implements Callable<Integer> {
   @Option(
     names = {"-p", "--progress"},
     description = "Print progress bars during conversion",
-    help = true,
     defaultValue = "false"
   )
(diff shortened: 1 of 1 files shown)


joshmoore:
_Uff_. My guess is that it was a copy-n-paste error.

chris-allan (commented):
Progress option interactions now work as I would expect. Can we use the standard options for `--help`?

 * https://picocli.info/#_mixin_standard_help_options

melissalinkert:
> Can we use the standard options for `--help`?
> 
> * https://picocli.info/#_mixin_standard_help_options

Main reason I didn't do that here was that we already define `-h` as the short option for setting the tile height.

PR #237, opened 2024-01-20
update Dockerfile

bgruening:
Seems that the latest version is not working with openjdk:8 anymore.
We got a: 

> Unrecognized option: --add-opens 
Error: Could not create the Java Virtual Machine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@
 
 # Similarly, the BUILD_IMAGE argument can be overwritten
 # but this is generally not needed.
-ARG BUILD_IMAGE=gradle:6.9-jdk8
+ARG BUILD_IMAGE=gradle:8.5-jdk8
 
 #
 # Build phase: Use the gradle image for building.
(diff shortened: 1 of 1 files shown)


melissalinkert:
Thanks for bringing this up, @bgruening.

As described in #239, removing support for JDK 8 was unintentional; my apologies for the inconvenience. The new 0.9.1 release should support JDK 8 again, so closing this as it should no longer be necessary. If you still see an issue with 0.9.1 though, please do let us know.

bgruening:
Thanks for fixing it. I'm happy to close this PR. But maybe it's useful to update gradle and openjdk - which on its own is deprecated?

sbesson:
@bgruening The Gradle version is definitely outdated as the build currently happens on Gradle 7.3.3 as defined in `gradle/wrapper`. As for the OpenJDK version, the library currently supports Java 8, 11 & 17 and there is currently no active plan to drop either of them.

The disconnect partly comes from the fact that none of us, certainly neither @melissalinkert nor myself, use the `Dockerfile` as part of our development workflow. This means the Dockerfile will go out-of-date and probably raises the wider question of whether there is still a rationale/motivation for maintaining this workflow in this repository or if it should be removed

Issue #175, opened 2022-11-04
Cannot convert svs to ome-tiff


gsakkis:
Hi, new to bioformats2raw/raw2ometiff. I'm trying to convert [CMU-1-Small-Region.svs](https://openslide.cs.cmu.edu/download/openslide-testdata/Aperio/CMU-1-Small-Region.svs) to OME-TIFF by first converting it to Zarr via bioformats2raw and then to OME-TIFF via raw2ometiff. However the latter complains about a missing directory:

 ```
$ bioformats2raw CMU-1-Small-Region.svs /tmp/CMU-1-Small-Region.ome.zarr
$ ll /tmp/CMU-1-Small-Region.ome.zarr
total 16
drwxrwxr-x 2 gsk gsk 4096 Νοε   4 18:32 OME
drwxrwxr-x 7 gsk gsk 4096 Νοε   4 18:32 0
drwxrwxr-x 4 gsk gsk 4096 Νοε   4 18:32 1
drwxrwxr-x 6 gsk gsk 4096 Νοε   4 18:32 2

$ raw2ometiff /tmp/CMU-1-Small-Region.ome.zarr /tmp/CMU-1-Small-Region.ome.tiff
Exception in thread "main" picocli.CommandLine$ExecutionException: Error while calling command (com.glencoesoftware.pyramid.PyramidFromDirectoryWriter@3b2cf7ab): java.lang.RuntimeException: java.io.IOException: Path '/tmp/CMU-1-Small-Region.ome.zarr/3' is not a valid path or not a directory.
```

What am I missing?

Versions:
```
$ bioformats2raw --version .
Version = 0.5.0
Bio-Formats version = 6.10.1
NGFF specification version = 0.4

$ raw2ometiff --version .
Version = 0.3.0
Bio-Formats version = 6.4.0
```

melissalinkert:
@gsakkis : That problem should have been fixed in https://github.com/glencoesoftware/raw2ometiff/pull/81, which was released in raw2ometiff 0.3.1. Can you try updating your raw2ometiff version to 0.3.1 and see if that works? You shouldn't need to re-run the bioformats2raw command, just raw2ometiff.

gsakkis:
@melissalinkert thanks for the prompt reply! raw2ometiff 0.3.1. doesn't seem to be available in [conda](https://anaconda.org/ome/raw2ometiff/files) that I use for installing the packages; I'll try to install it manually and see if it works.

[...]

PR #241, opened 2024-02-08
Update to Bio-Formats 7.2.0

melissalinkert:


sbesson (approved):
Looks good to me. Looking at open issues, it feels this is also addressing some of the remaining issues raised in https://github.com/glencoesoftware/bioformats2raw/issues/175. Should we link this PR to this issue? We might need to add a comment explaining that the layout using `bioformats2raw/raw2ometiff` will be most closely comparable to the one of `openslide` rather than `tifffile`.

PR #242, opened 2024-03-08
Metaxpress: get the correct number of planes for each series

melissalinkert:
...instead of assuming all series (wells/fields) have the same number of planes.


mabruce (approved):
OMERO import works as expected with this change.

PR #247, opened 2024-04-18
Fix a couple of compile warnings

melissalinkert:
Noticed while looking at #245. This is in the same spirit as https://github.com/ome/bioformats/pull/4177 (and other PRs linked from there).


sbesson (approved):
Tested the conversion of a few sample MIRAX files. Everything works as expected and the metadata is correctly populated in the `METADATA.ome.xml`

chris-allan (commented):
The semantics of `new Double()` are quite different than `DataTools.parseDouble()`. Was that also intended or desired here?

PR #251, opened 2024-05-20
Add "--no-original-metadata" option to omit original metadata from METADATA.ome.xml

### LINKED ISSUE #250: Add -no-sas option as exists in bfconvert
Would it be possible/easy to have the -no-sas option (as it exists in bfconvert) in bioformats2raw?

We're always fighting ways where PHI info sneaks into data that needs to become public. We want to switch to the bioformats2raw/raw2ometiff tools in favor of bfconvert but our SVS files have some problematic items in the metadata that ends up in the resulting ome.tif in SA's. So we want to keep the "regular" OME-XML metadata but not the SA stuff.

And I tried the --no-ome-meta-export option but then raw2ometiff is unhappy.

melissalinkert:
Fixes #250.

As the test demonstrates, converting input data that has original metadata with the new `--no-original-metadata` option should result in a `METADATA.ome.xml` that does not contain `OriginalMetadata` annotations. Default behavior should be unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,12 @@ which can then be used to create a hierarchy by date, username, and ID:
 Prevents the input file's OME-XML metadata from being saved.  There will no longer be an `OME` directory
 under the top-level output directory.
 
+#### --no-original-metadata
+
+Prevents `OriginalMetadata` annotations from being written to `OME/METADATA.ome.xml`.
+This can reduce the size of the OME-XML, and as discussed in [#250](https://github.com/glencoesoftware/bioformats2raw/issues/250),
+is one way to guard against unwanted experimental metadata being included in the conversion.
+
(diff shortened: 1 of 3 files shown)


dsudar:
I pulled in this PR, built, and ran a few tests on my data. All worked fine. Thumbs-up from me.

sbesson (approved):
Looks great and works as expected comparing the `METADATA.ome.xml` produced by  `bioformats2raw` with and without `--no-original-metadata`.

Is it worth also adding the option to the README? Otherwise, let's get this merged

melissalinkert:
> Is it worth also adding the option to the README? Otherwise, let's get this merged

Done in 76c993a.

sbesson (approved):

PR #256, opened 2024-09-23
Add Phenix INI file support

melissalinkert:
Similar to #213, just for Phenix v5.


sbesson (approved):
All looks good from a code review perspective. Passing onto @muhanadz for a round of functional testing

muhanadz (approved):
Tested the smaller phenix plate example we've looked at in our development tests. Had a few warnings for missing `OpenCV` libraries. Ran the conversion with `zlib` compressions as cblosc dependencies were missing.

Conversion with bioformat2raw was successful. Converted with with `raw2ometiff` and imported into OMERO successfully. LGTM.

PR #257, opened 2024-09-23
Fix `--fill-value` to use setFillColor(...) API

melissalinkert:
...so now this option can be used with all readers.

Noticed while working on #256 that `--fill-value` still only applied to `MiraxReader`. Given that `setFillColor` and `getFillColor` were added to `IFormatReader` in Bio-Formats 6.13.0 (see https://github.com/ome/bioformats/pull/3963), this updates `--fill-value` to make use of this new API. Several of the readers didn't make use of `getFillColor()` in `openBytes`, so that's been fixed too.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/MiraxReader.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/MiraxReader.java
@@ -131,7 +131,6 @@ public class MiraxReader extends FormatReader {
     JPEG2000CodecOptions.getDefaultOptions();
 
   private boolean fluorescence = false;
-  private Byte fillValue = null;
 
   private transient JPEGXRCodec jpegxrCodec = new JPEGXRCodec();
 
@@ -217,7 +216,7 @@ public class MiraxReader extends FormatReader {
 
     // set background color to black instead of the stored fill color
     // this is to match the default behavior of Pannoramic Viewer
-    Arrays.fill(buf, getFillValue());
+    Arrays.fill(buf, getFillColor());
 
     if (tileCache == null) {
       tileCache = CacheBuilder.newBuilder()
@@ -1057,9 +1056,10 @@ public class MiraxReader extends FormatReader {
    * Set the fill value for missing tiles.
    *
    * @param fill the fill value, or null to use the reader's default value
+   * @deprecated used setFillColor(Byte)
    */
   public void setFillValue(Byte fill) {
-    fillValue = fill;
+    setFillColor(fill);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/glencoesoftware/bioformats2raw/MiraxReader.java
```diff
@@ -1069,10 +1069,17 @@ public void setFillValue(Byte fill) {
    * for brightfield.
    *
    * @return fill value for missing tiles
+   * @deprecated use getFillColor()
```
sbesson, line 1072: As this is planned for `0.10.0`, is there a value in going through a deprecation rather than dropping these in favor of the new API defined upstream?

src/main/java/com/glencoesoftware/bioformats2raw/MiraxReader.java
```diff
@@ -1069,10 +1069,17 @@ public void setFillValue(Byte fill) {
...
    */
   public byte getFillValue() {
-    if (fillValue != null) {
-      return fillValue;
+    return getFillColor();
+  }
+
+  @Override
+  public Byte getFillColor() {
+    Byte color = getFillColor();
```
sbesson, line 1080: In my testing this caused a `java.lang.StackOverflowError: null` due to the infinite recursion

sbesson (changes requested):
Tested on a MRXS file with `--fill-value` set but the application entered an infinite recursion while calling `getFillColor`

src/main/java/com/glencoesoftware/bioformats2raw/MiraxReader.java
melissalinkert: Should be fixed with e3dd7ca.

src/main/java/com/glencoesoftware/bioformats2raw/MiraxReader.java
melissalinkert: I wasn't totally sure what the best approach would be here, so went with deprecating to be on the safer side. If we're all OK with just removing `getFillValue()` and `setFillValue()`, I'm fine with that - it would only impact anyone using `MiraxReader` directly, either by extending the reader or by calling those methods directly.

sbesson (approved):
Retested the last commit with a sample fluorescence MRXS dataset. This was converted into OME-Zarr via `bioformats2raw` using the default fill color and 3 values set via options `--fill-value=0`, `--fill-value=100` and `--fill-value=255`

Confirmed that the OME-Zarr display as expected with the default value still being set to 0 as per the modality

Screenshot 2024-10-07 at 09 30 14
Screenshot 2024-10-07 at 09 30 27

No objection to keeping the deprecation warnings for now and drop in a follow-up `0.x.0` release

Issue #254, opened 2024-08-06
Compression levels seem to have no effect on output file


FelipeMoser:
Hi, I've been playing around with bioformats2raw for a bit and I wanted to compare the different compression options available. I have some ome.tiffs that I'm converting to zarr.

However, I've noticed that the "--compression-properties" argument seems to have no effect on the compression itself outside of the values stored in the .zarray file. 

For example, if I set --compression=zlib and --compression-properties="level=1", the file size is exactly the same I get if I set --compression-properties="level=9". Similarly, using the default compression and using --compression-properties="clevel=1" or --compression-properties="clevel=9" results in the exact same file size. There's also no difference in computation time. In both cases, however, the .zarr file does change accordingly.

I'm using bioformats2raw version 0.9.1 

[edit] Adding to clarify, if I use zlib, the resulting file size is different than what I get if I use default (blosc). So the --compression argument does seem to work. The issue is only related to "--compression-properties".

melissalinkert:
Thanks for mentioning this, @FelipeMoser.

When you compare file sizes, are you comparing the actual number of bytes in each chunk file (e.g. `test.zarr/0/0/0/0/0/0/0`), or just the overall size in of the whole output directory? The difference between `level=1` (or `clevel=1` for blosc) and `level=9` (or `clevel=9`) can be pretty subtle, and is highly data-dependent, so may not be visible unless you're comparing the size in bytes of each file (and/or checksum).

For example, with bioformats2raw 0.9.3 and an artificial 512x512 image, convert the same input data with 4 different compression settings:

```
$ bin/bioformats2raw test.fake test-level1-zlib.zarr --compression zlib --compression-properties "level=1"
$ bin/bioformats2raw test.fake test-level9-zlib.zarr --compression zlib --compression-properties "level=9"
$ bin/bioformats2raw test.fake test-clevel1-blosc.zarr --compression-properties "clevel=1"
$ bin/bioformats2raw test.fake test-clevel9-blosc.zarr --compression-properties "clevel=9"
```


List the size in bytes of every file in both sets of zlib output, followed by the summary size of both outputs:
```
$ find test-*zlib.zarr -type f -exec ls -lgG '{}' \;
-rw-rw-r-- 1 33 Aug  6 13:11 test-level1-zlib.zarr/.zattrs
-rw-rw-r-- 1 591 Aug  6 13:11 test-level1-zlib.zarr/OME/METADATA.ome.xml
[...]

PR #262, opened 2024-10-17
Update readme to include section on compression options

melissalinkert:
As discussed with @mabruce in particular, and also mentioned in this week's PR review.

See also #254 and https://github.com/glencoesoftware/zarr2zarr/issues/9.


mabruce (approved):
Changes look good to me.

Minor point: `blosclz` and `lz4hc` are also valid cnames according to the JZarr docs. They are functional in my limited testing, though (with the default clevel) not magically better in the speed/space tradeoffs compared to zlib, blosc+lz4, or blosc+zstd.

sbesson (approved):
Thanks, this is really useful and can be amended as needed in the future.

PR #265, opened 2025-01-27
MRXS: fix position units and brightfield background color

melissalinkert:
As discussed on Friday with @erindiel. Expect to need a raw2ometiff PR as well for the same data, so will assign reviewers once both PRs are open.


erindiel (approved):
Confirmed with relevant datasets that:
- position units originally in pixels are now in physical units
- background color of black in a brightfield dataset is now white

PR #267, opened 2025-02-06
Update to Bio-Formats 8.1.0 and logback 1.3.15

melissalinkert:
See https://bio-formats.readthedocs.io/en/stable/about/whats-new.html#february


sbesson (approved):
Successfully tested a conversion on a public sample IMS file with LZ4 compression (which support was introduced in Bio-Formats 8.1.0)

PR #268, opened 2025-03-05
Update Bio-Formats to 8.1.1

melissalinkert:
See https://bio-formats.readthedocs.io/en/stable/about/whats-new.html#march


sbesson (approved):
Successfully tested with https://downloads.openmicroscopy.org/images/Imspector/zenodo-10476252/

PR #272, opened 2025-03-14
Replace non-existent `--compression-options` with `--compression-properties`

### LINKED ISSUE #271: compression-options vs compression-properties
In https://github.com/glencoesoftware/bioformats2raw/pull/262 some nice info was provided, but I can't get it to work with either 0.9.4 or 0.10.0-rc1.
Both don't appear to support `compression-options` and instead use `compression-properties`
Is there a plan to change the argument? or am I missing something?

melissalinkert:
Fixes #271.

Honestly not sure how I got this documentation so wrong in #262, but the intent is to keep the existing `--compression-properties` defined in https://github.com/glencoesoftware/bioformats2raw/blob/master/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java#L523.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ By default, output is compressed with Blosc using the `lz4` codec.
 
 To change the overall compression type, use `--compression <type>`. Supported types are `blosc`, `zlib`, and `null` (uncompressed).
 
-To change type-specific options, use `--compression-options <key=value>`.
+To change type-specific options, use `--compression-properties <key=value>`.
 
 Supported options for `blosc` are:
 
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
Using e.g. `-c blosc --compression-properties cname=zstd --compression-properties clevel=8` works for me!
Thanks!

psobolewskiPhD (approved):

README.md
```diff
@@ -337,7 +337,7 @@ The worker count should be set to 1 if the input data requires a Bio-Formats rea
 This is not a common case, but is a known issue with Imaris HDF data in particular.
 
 In general, expect to need to tune the above settings and measure
-relative performance. See the [Compression options section](#compression-options) above for more information on `--compression` and `--compression-options`.
+relative performance. See the [Compression options section](#compression-properties) above for more information on `--compression` and `--compression-properties`.
```
psobolewskiPhD, line 340: This link work work.
the heading in line 136 is still options.

psobolewskiPhD (changes requested):
the link line 340 won't work as it doesn't match line 136.
Either it needs to be reverted or the heading in 136 needs to be changed to `Compression Properties`

README.md
sbesson: Since the section discusses both available compressions and compression properties, my vote would be to keep the title as it is and fix the anchor link back to `#compression-option`

melissalinkert: Done in 1143228.

sbesson (approved):

psobolewskiPhD (approved):

Issue #276, opened 2025-06-06
Generation of invalid ome.zarr


jburel:
The option ``dimension-oder`` could lead to the generation of invalid ome.zarr
For example
```
bioformats2raw /path/to/file.mrxs /path/to/zarr-pyramid --dimension-order XYZTC
```
will have an zarray with order ``CTZYX``
which is invalid according to the ome.zarr spec
```
"the entries MUST be ordered by "type" where the "time" axis must come first (if present), followed by the "channel" or custom axis (if present) and the axes of type "space". If there are three spatial axes where two correspond to the image plane ("yx") and images are stacked along the other (anisotropic) axis ("z"), the spatial axes SHOULD be ordered as "zyx"."  https://ngff.openmicroscopy.org/0.4/index.html
```

PR #278, opened 2025-06-10
Deprecate `--dimension-order`

melissalinkert:
See #276.

As discussed with @jburel, @sbesson, and @dominikl, using `--dimension-order` can cause confusion as specifying anything other than `XYZCT` (the default) will produce data that is not valid according to the [current spec](https://ngff.openmicroscopy.org/0.5/index.html#multiscale-md). An alternate approach would be to just remove `--dimension-order` entirely, but we discussed that deprecating for now might be sufficient to reduce confusion without the risk of actually breaking anything.

This came up in the context of writing tests, not converting "real" data, so any thoughts on if/how this impacts existing non-standard conversion workflows would be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ This option is not present in 0.3.0 and later, as only Zarr output is supported.
 
 Versions 0.2.6 and prior used the input file's dimension order to determine the output
 dimension order, unless `--dimension-order` was specified.
-Version 0.3.0 and later uses the `TCZYX` order by default, for compatibility with https://ngff.openmicroscopy.org/0.2/#image-layout.
+Version 0.3.0 and later uses the `TCZYX` order by default, for compatibility with https://ngff.openmicroscopy.org/0.4/#image-layout.
 The `--dimension-order` option is considered deprecated and may be removed in a future release,
 as it results in invalid OME-NGFF data.
 


README.md
```diff
@@ -281,18 +281,9 @@ This option is not present in 0.3.0 and later, as only Zarr output is supported.
...
-    bioformats2raw /path/to/file.mrxs /path/to/zarr-pyramid --dimension-order XYCZT
-
-or can be set to use the input file's ordering, preserving the behavior of 0.2.6:
-
-    bioformats2raw /path/to/file.mrxs /path/to/zarr-pyramid --dimension-order original
-
-If a specific dimension order is passed to `--dimension-order`, it must be a valid dimension order as defined in
-the [OME 2016-06 schema](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Pixels_DimensionOrder).
-The specified dimension order is then reversed when creating Zarr arrays, e.g. `XYCZT` would become `TZCYX` in Zarr.
+Version 0.3.0 and later uses the `TCZYX` order by default, for compatibility with https://ngff.openmicroscopy.org/0.2/#image-layout.
```
sbesson, line 284:
```suggestion
Version 0.3.0 and later uses the `TCZYX` order by default, for compatibility with https://ngff.openmicroscopy.org/0.4/#image-layout.
```

PR #281, opened 2025-08-05
Do not store zero time increment values in coordinateTransforms

### LINKED ISSUE #280: Zero for coordinateTransforms scale
I'm trying to convert OME-TIFF images which are 2d whole slide image exported by the TissueGnostics microscope software. I wanted to use bioformats2raw to convert these to OME-Zarr for viewing in Neuroglancer. The conversion works but produces `.zattrs` metadata like this:

```
{
  "multiscales" : [ {
    "metadata" : {
      "version" : "Bio-Formats 7.3.0"
    },
    "axes" : [ {
      "unit" : "second",
      "name" : "t",
      "type" : "time"
    }, {
      "name" : "c",
      "type" : "channel"
    }, {
      "unit" : "micrometer",
      "name" : "z",
      "type" : "space"
    }, {
      "unit" : "micrometer",
      "name" : "y",
      "type" : "space"
    }, {
      "unit" : "micrometer",
      "name" : "x",
      "type" : "space"
    } ],
    "name" : "Slide 1 of 4_Region 002_channels.tiff",
    "datasets" : [ {
      "path" : "0",
      "coordinateTransformations" : [ {
        "scale" : [ 0.0, 1.0, 1.0, 0.335762932896614, 0.335762932896614 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "1",
      "coordinateTransformations" : [ {
        "scale" : [ 0.0, 1.0, 1.0, 0.671525865793228, 0.671525865793228 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "2",
      "coordinateTransformations" : [ {
        "scale" : [ 0.0, 1.0, 1.0, 1.343051731586456, 1.343051731586456 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "3",
      "coordinateTransformations" : [ {
        "scale" : [ 0.0, 1.0, 1.0, 2.686103463172912, 2.686103463172912 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "4",
      "coordinateTransformations" : [ {
        "scale" : [ 0.0, 1.0, 1.0, 5.372206926345824, 5.372206926345824 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "5",
      "coordinateTransformations" : [ {
        "scale" : [ 0.0, 1.0, 1.0, 10.744413852691649, 10.744413852691649 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "6",
      "coordinateTransformations" : [ {
        "scale" : [ 0.0, 1.0, 1.0, 21.488827705383297, 21.488827705383297 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "7",
      "coordinateTransformations" : [ {
        "scale" : [ 0.0, 1.0, 1.0, 42.977655410766594, 42.977655410766594 ],
        "type" : "scale"
      } ]
    } ],
    "version" : "0.4"
  } ],
  "omero" : {
    "channels" : [ {
      "color" : "0000FF",
      "coefficient" : 1,
      "active" : true,
      "label" : "DAPI",
      "window" : {
        "min" : 0.0,
        "max" : 65535.0,
        "start" : 0.0,
        "end" : 65535.0
      },
      "family" : "linear",
      "inverted" : false
    }, {
      "color" : "FF8000",
      "coefficient" : 1,
      "active" : true,
      "label" : "CY3",
      "window" : {
        "min" : 0.0,
        "max" : 65535.0,
        "start" : 0.0,
        "end" : 65535.0
      },
      "family" : "linear",
      "inverted" : false
    }, {
      "color" : "FF00FF",
      "coefficient" : 1,
      "active" : true,
      "label" : "CY5",
      "window" : {
        "min" : 0.0,
        "max" : 65535.0,
        "start" : 0.0,
        "end" : 65535.0
      },
      "family" : "linear",
      "inverted" : false
    } ],
    "rdefs" : {
      "defaultT" : 0,
      "model" : "color",
      "defaultZ" : 0
    }
  }
}
```

This fails to load in Neuroglancer with the following error:
"Expected positive finite floating-point number, but received: 0". 

This is because the coordinate transforms collapse the time axis to 0 for every scale level. If I change the time axes from 0.0's to 1.0's it loads without issue. 

Is there any way to get around this issue? Is the bug in bioformats2raw, Neuroglancer, or TissueGnostics?

sbesson:
Fixes #280

While the OME model allows any floating point value for the time increment, a 0 value in the scale transformation in the coordinateTransforms can create confusion when interpreted as a downsampling factor between resolutions

Using the sample file provided in the corresponding issue, `bioformats2raw` should produce an OME-Zarr dataset where the `scale` transformation includes a `0.0` value for the time axis.
With this change included, PixelsTimeIncrement values of zero should be ignored and the `scale` arrays should use the default `1.0`

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -2681,7 +2681,13 @@ public class Converter implements Callable<Integer> {
       case 'z':
         return meta.getPixelsPhysicalSizeZ(seriesIndex);
       case 't':
-        return meta.getPixelsTimeIncrement(seriesIndex);
+        Quantity timeIncrement = meta.getPixelsTimeIncrement(seriesIndex);
+        if (timeIncrement != null && timeIncrement.value().doubleValue() > 0) {
+          return meta.getPixelsTimeIncrement(seriesIndex);
+        }
+        else {
+          return null;
+        }
       default:
         return null;
     }


src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -2681,7 +2681,13 @@ private Quantity getScale(
       case 'z':
         return meta.getPixelsPhysicalSizeZ(seriesIndex);
       case 't':
-        return meta.getPixelsTimeIncrement(seriesIndex);
+        Quantity timeIncrement = meta.getPixelsTimeIncrement(seriesIndex);
+        if (timeIncrement != null && timeIncrement.value().doubleValue() > 0) {
+          return meta.getPixelsTimeIncrement(seriesIndex);
```
melissalinkert, line 2686: Any reason not to just `return timeIncrement` here instead of calling the getter again?

melissalinkert (commented):
Makes sense to me (nitpicky comment aside :smile:).

Separately, I am a little surprised that setting 0 was allowed, and wonder why `TimeIncrement` is `xsd:float` in OME-XML schema in the first place. All of the other units are `PositiveFloat`, and having `TimeIncrement` as `PositiveFloat` as well would have avoided this problem in the first place. I don't remember and can't easily find a reference to any discussion, but it would be interesting to know if there is a use case for non-positive `TimeIncrement` values that might have motivated that type difference. Or maybe it's just that time is the only non-physical dimension with units? In any case, that's outside the scope of this pull request.

PR #289, opened 2025-09-02
Restrict units to what is defined in NGFF schema

### LINKED ISSUE #288: Unit "pixel" propagated to NGFF axes causes Neuroglancer parse error
I used Fiji to convert the example found under "Open Samples->MRI Stack" to OME-TIFF. That image can be found here: https://s3.janelia.org/scicompsoft/examples/

This TIFF has metadata like this: `PhysicalSizeXUnit="pixel"`

When I used bioformats2raw to convert that TIFF into a OME-Zarr, it propagated the "pixel" unit to the axes like so:

```
    "axes" : [ {
      "name" : "t",
      "type" : "time"
    }, {
      "name" : "c",
      "type" : "channel"
    }, {
      "unit" : "pixel",
      "name" : "z",
      "type" : "space"
    }, {
      "unit" : "pixel",
      "name" : "y",
      "type" : "space"
    }, {
      "unit" : "pixel",
      "name" : "x",
      "type" : "space"
    } ],
``` 

This then causes Neuroglancer to throw an error when loading the image:
> Error parsing "axes" property: Error parsing "unit" property: Unsupported unit: "pixel"

The unit "pixel" is not listed in the [allowable units in the NGFF spec](https://ngff.openmicroscopy.org/latest/#axes-md). Since the units are a SHOULD and not a MUST, it seems that technically "pixel" is allowed, but in practice this can create downstream issues. 

FWIW, deleting the units makes Neuroglancer work. On the other hand, the "pixel" unit is accepted or ignored by other viewers like Vizarr and Vol-E. 

One thing that could be done here is to simply leave the unit out when it does not conform to the list in the spec. Alternatively, maybe Neuroglancer and other consumers simply need to be more permissive in parsing. What do you all think?

melissalinkert:
Fixes #288.

This lists valid space and time units per https://ngff.openmicroscopy.org/latest/#axes-md, and rejects anything not included in those lists. As indicated in #144, `pixel` is one of several units that are supported by OME-XML schema but not NGFF.

I debated whether to implement this by defining NGFF schema units, or defining just units that don't overlap between the two schemas (as in https://github.com/glencoesoftware/bioformats2raw/pull/144#issuecomment-1087806848). I think the former approach is going to be easier to keep up to date, but happy to hear other opinions.

With the test data mentioned in #288, `bioformats2raw --debug mri.ome.tiff mri.zarr` should log that `pixel` units are being omitted, and `mri.zarr/0/.zattrs` should show no `unit` attribute for anything under `axes`.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -127,6 +127,24 @@ public class Converter implements Callable<Integer> {
   /** NGFF specification version.*/
   public static final String NGFF_VERSION = "0.4";
 
+  /** Units allowed by NGFF specification for length/space axes. */
+  private static final List<String> LENGTH_UNITS = Arrays.asList(
+    "angstrom", "attometer", "centimeter", "decimeter", "exameter",
+    "femtometer", "foot", "gigameter", "hectometer", "inch", "kilometer",
+    "megameter", "meter", "micrometer", "mile", "millimeter", "nanometer",
+    "parsec", "petameter", "picometer", "terameter", "yard", "yoctometer",
+    "yottameter", "zeptometer", "zettameter"
+  );
+
+  /** Units allowed by NGFF specification for time axes. */
+  private static final List<String> TIME_UNITS = Arrays.asList(
+    "attosecond", "centisecond", "day", "decisecond", "exasecond",
+    "femtosecond", "gigasecond", "hectosecond", "hour", "kilosecond",
+    "megasecond", "microsecond", "millisecond", "minute", "nanosecond",
+    "petasecond", "picosecond", "second", "terasecond", "yoctosecond",
+    "yottasecond", "zeptosecond", "zettasecond"
+  );
+
   private volatile Path inputPath;
   private volatile String outputLocation;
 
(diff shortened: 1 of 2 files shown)


sbesson (changes requested):
Using `bioformats2raw 0.10.1`, I was able to reproduce the original issue either using the Fiji workflow described in #288 (Open the MRI sample and export as OME-TIFF) or a synthetic sample file like `test&physicalSizeX=1pixel&physicalSizeY=1pixel.fake`. In both cases, the conversion generates a Zarr dataset where the unit of the x and y the physical axes are expressed in `pixel`.

With a build of `bioformats2raw` including this PR, the Zarr datasets generated by the conversion are binary identical except for the physical axes which are now unitless

```diff
sbesson@Sebastien-GS-MacBook-Pro-2025 bioformats2raw % diff -ur test.zarr test_pr289.zarr 
diff -ur test.zarr/0/.zattrs test_pr289.zarr/0/.zattrs
--- test.zarr/0/.zattrs	2025-09-04 08:46:45
+++ test_pr289.zarr/0/.zattrs	2025-09-04 08:54:26
@@ -14,11 +14,9 @@
       "name" : "z",
       "type" : "space"
     }, {
-      "unit" : "pixel",
       "name" : "y",
       "type" : "space"
     }, {
-      "unit" : "pixel",
       "name" : "x",
       "type" : "space"
     } ],
sbesson@Sebastien-GS-MacBook-Pro-2025 bioformats2raw % diff -ur mri.zarr mri_pr289.zarr   
diff -ur mri.zarr/0/.zattrs mri_pr289.zarr/0/.zattrs
--- mri.zarr/0/.zattrs	2025-09-04 08:45:28
+++ mri_pr289.zarr/0/.zattrs	2025-09-04 08:54:19
@@ -11,15 +11,12 @@
       "name" : "c",
       "type" : "channel"
     }, {
-      "unit" : "pixel",
       "name" : "z",
       "type" : "space"
     }, {
-      "unit" : "pixel",
       "name" : "y",
       "type" : "space"
     }, {
-      "unit" : "pixel",
       "name" : "x",
       "type" : "space"
     } ],
```

The proposed changes are consistent with the discussion in https://github.com/glencoesoftware/bioformats2raw/pull/144 and feels like a reasonable approach given the behavior of other tools in the NGFF ecosystem. /cc @ome/ngff

Could we add a unit test using fake files to capture the expected behavior?

melissalinkert:
> The proposed changes are consistent with the discussion in https://github.com/glencoesoftware/bioformats2raw/pull/144 and feels like a reasonable approach given the behavior of other tools in the NGFF ecosystem. /cc @ome/ngff

Looks like the team mention might not have worked (maybe due to different organization?) - cc @joshmoore.

> Could we add a unit test using fake files to capture the expected behavior?

Done in 4b39f7e.

PR #302, opened 2026-01-29
Switch from jzarr to zarr-java for v2/NGFF 0.4

melissalinkert:
This fully removes the dependency on jzarr, and uses zarr-java for all versions.

There are a few test failures here, which are specific to v2 plate data. See note in `Converter` line 2816. Since there have been a few relevant fixes in zarr-java since 0.0.9 (especially https://github.com/zarr-developers/zarr-java/pull/36), I expect to keep this as a draft until a new zarr-java version is released.

There are also a few commented to-do items around compression properties (likely needs documentation in addition to code changes), and testing of group key counts.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ dependencies {
     implementation 'ome:formats-gpl:8.3.0'
     implementation 'info.picocli:picocli:4.7.5'
     implementation 'com.univocity:univocity-parsers:2.8.4'
-    implementation 'dev.zarr:zarr-java:0.0.10'
+    implementation 'dev.zarr:zarr-java:0.1.0'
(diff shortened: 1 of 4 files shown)


melissalinkert:
> I see that each multiscales directory contains a zarr.json and a directory named `c`. The specification states that `Each multiscale level is stored as a separate Zarr array, which is a folder containing chunk files which compose the array.` So is `c` referring to "chunk"?

The extra `c` directory is in the v3 spec (not something we've chosen in bioformats2raw):

https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#id36

I couldn't find a reference to what that is meant to signify, but also assumed "chunk".

PR #305, opened 2026-03-13
MCD: swap Channel.Name and Channel.Fluor

melissalinkert:
As discussed with @muhanadz / @erindiel / @mabruce / @heather-ashmore.

This just swaps which value goes into the `Name` attribute and which attribute goes into `Fluor`.


mabruce:
This change makes sense to me, as I'd expect to see the protein target in the channel name and metal ion in the "fluorophore" metadata.

PR #315, opened 2026-04-15
OME-Zarr 0.5: add version to well group attributes

sbesson:
Discovered while generating a set of OME-Zarr 0.5 datasets with 0.12.0-rc3, the metadata written for the well group is currently invalid as it is missing a `version` attribute - see e..g https://ome.github.io/ome-ngff-validator/?source=https://gs-public-zarr-dev.s3.amazonaws.com/zarrv3/NIRHTa%252B001.zarr/A/1/

This PR fixes this by:
- populating the attribute storing to include the version as for the other groups
- extending the HCS tests to check the well group attributes: version and well/images

The README is also expanded to cover the multi-version support and the `--ngff-version` option. As this is slightly outside the scope of the initial problem, happy to move the last commit to a dedicated PR if that is preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,14 +221,19 @@ Output Formatting Options
 =========================
 
 By default, the output of `bioformats2raw` will be a
-[Zarr dataset](https://zarr.readthedocs.io/en/stable/spec/v2.html) which follows the
+[Zarr dataset](https://zarr-specs.readthedocs.io/en/latest/specs.html) which follows the
 metadata conventions defined by the
-[OME-NGFF 0.4 specification](https://ngff.openmicroscopy.org/0.4/) including the
-[bioformats2raw.layout specification](https://ngff.openmicroscopy.org/0.4/#bf2raw).
+[version 0.4](https://ngff.openmicroscopy.org/specifications/0.4/) of the OME-Zarr specification including the
+[bioformats2raw.layout specification](https://ngff.openmicroscopy.org/specifications/0.4/#bioformats2raw-layout-transitional).
 
 Several formatting options can be passed to the converter and will result in a Zarr dataset
 that is not compatible with raw2ometiff and does not strictly follow the OME-NGFF
 specification but may be suitable for other applications.
+
+#### --ngff-version
+
+Specifies the version of the [OME-Zarr specification](https://ngff.openmicroscopy.org/specifications/index.html)
+that should be used while writing Zarr. Current supported values are 0.4 and 0.5.
 
 #### --pyramid-name
 
--- a/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
@@ -2752,6 +2752,7 @@ public class Converter implements Callable<Integer> {
           wellMap.put("images", imageList);
           Attributes columnAttrs = new Attributes();
           columnAttrs.put("well", wellMap);
+          columnAttrs.put("version", getNGFFVersion().toString());
 
           String rowPath = index.getRowPath();
           String columnPath = index.getColumnPath();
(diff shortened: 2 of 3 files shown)


src/main/java/com/glencoesoftware/bioformats2raw/Converter.java
```diff
@@ -2752,6 +2752,7 @@ private void saveHCSMetadata(IMetadata meta)
           wellMap.put("images", imageList);
           Attributes columnAttrs = new Attributes();
           columnAttrs.put("well", wellMap);
+          columnAttrs.put("version", getNGFFVersion().toString());
```
melissalinkert: This will include the `version` attribute for 0.4 as well - is it worth moving this line under the `getV3()` check in line 2760?

sbesson: 👍 [c0327fe](https://github.com/glencoesoftware/bioformats2raw/pull/315/commits/c0327fe941a1e5e49533f82a5a15f04aee5cefa9)

melissalinkert (approved):

README.md
```diff
@@ -221,15 +221,20 @@ Output Formatting Options
...
+[Zarr dataset](https://zarr-specs.readthedocs.io/en/latest/specs.html) which follows the
 metadata conventions defined by the
-[OME-NGFF 0.4 specification](https://ngff.openmicroscopy.org/0.4/) including the
-[bioformats2raw.layout specification](https://ngff.openmicroscopy.org/0.4/#bf2raw).
+[version 0.4](https://ngff.openmicroscopy.org/specifications/0.4/) of the OME-Zarr specification including the
+[bioformats2raw.layout specification](https://ngff.openmicroscopy.org/specifications/0.4/#bioformats2raw-layout-transitional).
 
 Several formatting options can be passed to the converter and will result in a Zarr dataset
 that is not compatible with raw2ometiff and does not strictly follow the OME-NGFF
 specification but may be suitable for other applications.
```
erindiel, line 231: I was considering changing this to `may result` and `may not strictly follow`, but I think instead the note in `--ngff-version` section is better.

README.md
```diff
@@ -221,15 +221,20 @@ Output Formatting Options
...
 
+#### --ngff-version
+
+Specifies the version of the [OME-Zarr specification](https://ngff.openmicroscopy.org/specifications/index.html)
+that should be used while writing Zarr. Current supported values are 0.4 and 0.5.
+
```
erindiel, line 237:
```suggestion
Specifies the version of the [OME-Zarr specification](https://ngff.openmicroscopy.org/specifications/index.html)
that should be used while writing Zarr. Current supported values are 0.4 and 0.5.
Both options produce outputs which are compatible with raw2ometiff.

```

Are we comfortable saying this? Assuming both will be released together...

erindiel (approved):
Confirmed with conversion of https://downloads.openmicroscopy.org/images/PerkinElmer-Operetta/idr0034/ with `--ngff-version 0.5` that the zarr.json in each well (`.zarr/<letter>/<number>/zarr.json`) contained a `version` key:

```
{
  "zarr_format" : 3,
  "node_type" : "group",
  "attributes" : {
    "ome" : {
      "well" : {
        "images" : [ {
          "path" : "0",
          "acquisition" : 0
        }, {
          "path" : "1",
          "acquisition" : 0
        }, {
          "path" : "2",
          "acquisition" : 0
        }, {
          "path" : "3",
          "acquisition" : 0
        }, {
          "path" : "4",
          "acquisition" : 0
        }, {
          "path" : "5",
          "acquisition" : 0
        }, {
          "path" : "6",
          "acquisition" : 0
        }, {
          "path" : "7",
          "acquisition" : 0
        }, {
          "path" : "8",
          "acquisition" : 0
        } ]
      },
      "version" : "0.5"
    }
  }
}
```

I have a small note on the README, but it isn't required for this to be merged.

README.md
sbesson: I think this is correct if no additional option is passed but as noted above if additional formatting options are passed, the Zarr dataset might be stored using a layout that makes it incompatible with raw2ometiff or other OME-Zarr aware tooling

melissalinkert:
Merging as discussed with @Yajing826 and @sbesson so that we can unblock adding tests to #317. Propose that we do a final review of the README next week prior to tagging 0.12.0 to catch any final documentation issues.

PR #331, opened 2026-06-30
Add note about `--max-cached-tiles` option

melissalinkert:
Prompted by some recent conversations around .mrxs conversion, I realized we have nothing in the README that describes the `--max-cached-tiles` option. cc @muhanadz, @mabruce


erindiel (approved):
This reads clearly to me and reflects the performance impact reported by @muhanadz

sbesson (approved):

PR #332, opened 2026-07-15
Improve converter and cli

Darlokt:
Hej,

while I was setting up the conda builds, I found some weird behavior with the exit codes (with `--version` in particular with `-1`/ `255`, leading to some funny build logs), which were also out of match with `raw2ometiff`/Unix standard. This led me a bit down the rabbit hole of the converter's error handling, output streams and exit codes, and I ended up making a number of changes to make the CLI more predictable and safer. It now validates invalid settings before conversion starts, propagates errors that were previously swallowed, protects existing output during `--overwrite`, and gives the CLI consistent output streams and exit statuses.

It also adds focused CLI regression tests and removes obsolete converter code.

Several failure paths could previously continue with invalid values, return before reporting an asynchronous error, or modify an existing output before the input was fully opened and validated. The CLI also mixed informational and diagnostic output and did not consistently distinguish usage errors, operational failures, and interruption.

Together, this made automation less reliable and could leave users without a clear indication that conversion had failed, was rejected etc.

### Validate converter configuration early

- Add `InvalidConfigurationException` for invalid CLI and programmatic settings.
- Reject non-positive tile, chunk, shard, worker, resolution, and target-size values.
- Validate cache size, fill value, log level, reader-option syntax, output options, and extra reader classes.
- Validate compression properties, codecs, levels, and Zarr-version compatibility.
- Reject negative, duplicate, and out-of-range series selections.
- Validate scale format strings and additional CSV arguments before writing output.
- Enforce compatibility between unnested output and NGFF version 0.1.

### Propagate conversion failures reliably

- Propagate reader initialization, metadata service, series, and asynchronous chunk failures to callers.
- Correctly unwrap unexpected asynchronous exceptions.
- Stop worker tasks safely when conversion is interrupted and restore the thread interrupt flag.
- Close all initialized readers during failure cleanup.
- Remove newly created memo files with checked filesystem operations and report cleanup failures as concise warnings.
- Fall back to pixel-type ranges when `OMERO` min/max calculation is unavailable.

### Make overwrite handling safer

- Open the input and complete input-dependent validation before modifying existing output.
- Preserve existing output when input or configuration validation fails.
- Reject overwrite requests on remote filesystems where safe deletion is unsupported.
- Delete local output paths in reverse order using checked operations, closing the `Files.walk` stream correctly.
- Stop immediately when an output path cannot be removed.
- Track whether output modification has begun so the CLI only warns about potentially incomplete output when appropriate.

### Standardize CLI behavior

- Add explicit `picocli` command metadata and version reporting.
- Send help and version output to standard output.
- Send logs, warnings, progress, and errors to standard error.
- Report routine errors concisely without a stack trace, retain stack traces when debug logging is enabled.
- Use stable exit statuses, in line with common Unix conventions, to distinguish between success, input/output failures, configuration errors, and interruption:

  | Status | Meaning |
  | ---: | --- |
  | `0` | Success, help, or version output |
  | `1` | Input, output, dependency, or conversion failure |
  | `2` | Command-line or configuration error |
  | `130` | Interrupted conversion |

- Warn when a failed or interrupted conversion may have left incomplete output.

### Tests and cleanup

- Add CLI regression coverage for output streams, exit statuses, invalid configuration, overwrite safety, partial output, and interruption.
- Remove unused converter state and the obsolete stopwatch helper.

I think especially in the context of cleanup and error handling, there can still a lot be done, but this I think is a good start to make the CLI more predictable and safer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,14 +80,29 @@ You can configure logging by editing the provided `logback.xml` or by specifying
     JAVA_OPTS="-Dlogback.configurationFile=/path/to/external/logback.xml" \
     bioformats2raw ...
 
-Alternatively you can use the `--debug` flag, optionally writing the stdout to a file:
+Alternatively you can use the `--debug` flag, optionally writing the log output to a file:
 
-    bioformats2raw /path/to/file.mrxs /path/to/zarr-pyramid --debug > bf2raw.log
+    bioformats2raw /path/to/file.mrxs /path/to/zarr-pyramid --debug 2> bf2raw.log
 
 The `--log-level` option takes an [slf4j logging level](https://www.slf4j.org/faq.html#fatal) for additional simple logging configuration.
 `--log-level DEBUG` is equivalent to `--debug`. For even more verbose logging:
 
     bioformats2raw /path/to/file.mrxs /path/to/zarr-pyramid --log-level TRACE
+
+Command output and exit status
+==============================
+
+Help and version information are written to standard output. Logging,
+progress bars, warnings, and errors are written to standard error. A successful
+conversion does not otherwise write to standard output.
+
+Routine errors are concise. Use `--debug` or `--log-level DEBUG` to include a
+Java stack trace for troubleshooting.
+
+If conversion fails or is interrupted after writing starts, the output is kept
+and may be incomplete. When `--overwrite` is used, input is checked before the
+existing local output is removed. Each old path is deleted with error checking,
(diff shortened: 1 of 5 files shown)


melissalinkert:
This is a significant rewrite of existing logging, error, and exception handling. Changes of this scope and magnitude are at odds with our current priorities, and have consequences for existing tools and workflows that make use of bioformats2raw. We are unfortunately not able to accept this change.

PR #337, opened 2026-07-31
Remove reference to conda

melissalinkert:
As discussed with @sbesson and @chris-allan, the reference to conda is being removed from the README to avoid confusion, since we do not maintain conda packages.

This is a documentation change only, so no testing is required (FYI @erindiel).

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,9 @@ For additional information, please see:
 Installation
 ============
 
-1. Download and unpack a release artifact:
+Download and unpack a release artifact:
 
     https://github.com/glencoesoftware/bioformats2raw/releases
-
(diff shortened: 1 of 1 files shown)


sbesson (approved):
This is consistent with the discussion we had with Melissa and Chris. This repository contains the sources for the library library and offers zip bundles generated by the build system as release artifacts. These artifacts can be downloaded, packaged and distributed using various mechanisms (Docker, Conda, Singularity...) but we are not maintaining any of these third-party distributions.

In the case of Conda, the OME recipe is currently transitioning between channels and there are alternative packages like the `batchconvert` recipe from Euro-BioImaging. We are removing this section to reduce the requirement to keep this README up-to-date with downstream decisions.

PR #335, opened 2026-07-16
Write anatomical orientation if available (RFC-4)

melissalinkert:
Closes #329. This depends upon an open Bio-Formats pull request (https://github.com/ome/bioformats/pull/4454), so is a draft for now. This will also need to make use of the `1.0-DEV` version support and tests added in #330, as RFC-4 support should only be written with `--ngff-version 1.0-DEV`. I just didn't want to introduce conflicting commits here.

As noted in #329, the general idea is that Bio-Formats proper implements most of RFC-4, including the controlled vocabulary therein. This pull request simply consumes axis orientations provided by Bio-Formats. The API defined in Bio-Formats and consumed here should be flexible enough to handle non-anatomical orientations, if that were to be added in a future RFC.

Basic tests make use of `FakeReader`, but Analyze, Nifti, MINC, Bruker, Varian, and DICOM are all candidates for testing once we have suitable reference data (see https://github.com/glencoesoftware/bioformats2raw/issues/329#issuecomment-4861441277).

I expect the build to fail here due to the need for https://github.com/ome/bioformats/pull/4454. To test locally, check out that Bio-Formats branch, then `mvn clean package install`, then `./gradlew clean build` on this pull request's branch.

PR #338, opened 2026-08-25
Add reader for Yokogawa CQ1/CQ3000 data

melissalinkert:
This is a variant of OME-TIFF represents a plate. There is a main OME-TIFF file (with an extra image that needs to be removed), and then separate linked TIFF files that contain the actual pixel data. Channel names are overridden as discussed with @mabruce.

This depends upon changes in https://github.com/ome/bioformats/pull/4483, and as such the build will currently fail. To test locally, check out the Bio-Formats branch and run `mvn clean package install`, then build this branch (c71e068 allows the installed snapshot to be picked up).

Similar to #335, I expect we'll want to get https://github.com/ome/bioformats/pull/4483 merged and tagged in an RC, then update this to consume the RC before merging.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/CQ1Reader.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/CQ1Reader.java
@@ -33,6 +33,8 @@ import ome.xml.model.primitives.NonNegativeInteger;
 public class CQ1Reader extends OMETiffReader {
 
   private List<String> extraFiles = new ArrayList<String>();
+  private transient boolean addPlateAcquisition = false;
+  private transient boolean addPlateName = false;
 
   /** Construct a new CQ1 reader. */
   public CQ1Reader() {
@@ -62,6 +64,8 @@ public class CQ1Reader extends OMETiffReader {
     super.close(fileOnly);
     if (!fileOnly) {
       extraFiles.clear();
+      addPlateAcquisition = false;
+      addPlateName = false;
     }
   }
 
@@ -91,12 +95,14 @@ public class CQ1Reader extends OMETiffReader {
 
     // set the plate name to something useful
 
-    String[] files = getSeriesUsedFiles();
-    Location pixels = new Location(files[files.length - 1]).getAbsoluteFile();
-    pixels = pixels.getParentFile();
-    String pixelsPath = pixels.getName();
-    String parentDirectory = pixels.getParentFile().getName();
-    metadataStore.setPlateName(parentDirectory + "_" + pixelsPath, 0);
+    if (addPlateName) {
+      String[] files = getSeriesUsedFiles();
+      Location pixels = new Location(files[files.length - 1]).getAbsoluteFile();
+      pixels = pixels.getParentFile();
+      String pixelsPath = pixels.getName();
+      String parentDirectory = pixels.getParentFile().getName();
+      metadataStore.setPlateName(parentDirectory + "_" + pixelsPath, 0);
+    }
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


mabruce (approved):
Tested alongside the linked PRs, I was able to successfully convert a CQ3000 plate to OME-Zarr (v0.4) and import into OMERO Plus. The channel names were as expected.

src/main/java/com/glencoesoftware/bioformats2raw/CQ1Reader.java
```diff
@@ -0,0 +1,337 @@
...
+  private static final String MEASUREMENT_PROTOCOL = "MeasurementProtocol.xml";
+
+  private List<String> extraFiles = new ArrayList<String>();
+  private transient boolean addPlateAcquisition = false;
+  private transient boolean addPlateName = false;
+  private transient String protocolFile = null;
+
+  /** Construct a new CQ1 reader. */
+  public CQ1Reader() {
+    super("Yokogawa CQ1", new String[] {"ome.tif", "ome.tiff"});
```
sbesson: Should this be updated to represent the fact this reader is used for CQ1 & CQ3000 instruments?

sbesson (changes requested):
Yokogawa CQ1/CQ300 instruments store acquired data using a file format based on the OME-TIFF specification. Over the years, we have received several examples of such datasets including a few public ones - see https://zenodo.org/records/4751185 and https://zenodo.org/records/5727645.
This collection allowed us to identify several bespoke usages or deviations from the official specification including but not limited to:

- the presence of a single orphaned `Image` alongside `Image` elements in a HCS structure, see https://forum.image.sc/t/problems-in-omero-with-datasets-from-a-yokogawa-cq1-microscope/59880 . Mixed HCS/non HCS data is valida according to the OME data model but is incompatible with the current supported versions the OME-Zarr specification (0.4 & 0.5)
- the association of every `WellSample` with a different `PlateAcquisition` element - see https://forum.image.sc/t/yokogawa-cq1-in-omero-a-lot-of-plateacquisition/90017
- the presence of `Pixels` elements with no `TiffData`, `BinData` or `MetadataOnly` elements, which is invalid as per the OME model
- the storage of additional acquisition metadata like channel names outside the OME-XML

In practice, this means these datasets generated by the Yokogawa instruments might be minimally readable by OMETiffReader. However, the strict interpretation of the OME-XML metadata makes this reader insufficient or limited for many downstream purposes.

As discussed in https://github.com/ome/bioformats/pull/4483#pullrequestreview-5064556216, adding support proprietary OME-TIFF variants is not an option we are considering in Bio-Formats, both from a support but also from a philosophy perspective.
Instead, this PR proposes a conversion-only workflow allowing to transform datasets generated by CQ instruments into open formats like OME-Zarr or OME-TIFF. This is achieved by adding an extra reader derived from OMETiffReader which handle all the variants described above and can read and populate extra metadata from companion files.

The reader has been extensively tested in the the Glencoe products and has been refined over the years with the knowledge from new datasets. Commits [9ceaab8](https://github.com/glencoesoftware/bioformats2raw/pull/338/commits/9ceaab8efc01a7f236e845cb537d1cf56252ed10), [72b1d56](https://github.com/glencoesoftware/bioformats2raw/pull/338/commits/72b1d569242a9b18f125fc33b5284b6f6f3436ce) and [6c3bd15](https://github.com/glencoesoftware/bioformats2raw/pull/338/commits/6c3bd15ead5f5fab5d67da2d831ac1417cdfa817) extend the reader to support datasets generated by  Yokogawa CQ3000 instruments and parse acquisition metadata from the MeasurementProcotol XML file. The only blocker before merging this it to bump the Bio-Formats version to consume the first 9.0.0 release candidate version which will include https://github.com/ome/bioformats/pull/4483.

src/main/java/com/glencoesoftware/bioformats2raw/CQ1Reader.java
melissalinkert: Done in 46b4fe5, to reflect what is in `isThisType`. That also raises the question of whether the reader class itself should be renamed. I did not do that, but let me know if `CQReader` (or something else) would be better.

sbesson: I considered this while https://github.com/glencoesoftware/bioformats2raw/pull/338#discussion_r3907600965.

My biggest concern is that we might create additional maintenance burden for ourselves while adding no value for the end-user. We also have similar precedent with CV7000Reader which reads the file format that is written by both CV7000 and CV8000 acquisition systems. Right now, I feel the value of renaming the reader is low compared to its cost.

sbesson (approved):
Retested the latest build of `bioformats2raw` with the release candidate of Bio-Formats 9.0.0-rc1 against sample CQ1/CV8000 plates, converting them into OME-Zarr 0.5 datasets and importing them into OMERO Plus.

Everything works as expected and the introspection of the metadata at the Zarr multiscales group confirms the correct version of Bio-Formats is used.

From the documentation perspective, we might need want to add the new reader to the list in https://github.com/glencoesoftware/bioformats2raw/blob/master/README.md#additional-readers but this is not a blocker for including this work.

PR #340, opened 2026-09-11
CQ1 reader: set brightfield channel color to white

melissalinkert:
Without this change, the `Color` on `Channel` in `OME/METADATA.ome.xml` was blank for `Brightfield` channels. The `color` attribute of the corresponding channel under `omero` in the Zarr metadata was set to red due to https://github.com/glencoesoftware/bioformats2raw/blob/master/src/main/java/com/glencoesoftware/bioformats2raw/Colors.java#L117 and https://github.com/glencoesoftware/bioformats2raw/blob/master/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java#L3287.

With this change, OME-XML and Zarr metadata should agree that the `Brightfield` channel is white.

### DIFF
--- a/src/main/java/com/glencoesoftware/bioformats2raw/CQ1Reader.java
+++ b/src/main/java/com/glencoesoftware/bioformats2raw/CQ1Reader.java
@@ -28,6 +28,7 @@ import ome.xml.model.Image;
 import ome.xml.model.Plate;
 import ome.xml.model.Well;
 import ome.xml.model.WellSample;
+import ome.xml.model.primitives.Color;
 import ome.xml.model.primitives.NonNegativeInteger;
 
 import org.w3c.dom.Element;
@@ -46,6 +47,7 @@ public class CQ1Reader extends OMETiffReader {
     LoggerFactory.getLogger(CQ1Reader.class);
 
   private static final String MEASUREMENT_PROTOCOL = "MeasurementProtocol.xml";
+  private static final String BRIGHTFIELD_NAME = "Brightfield";
 
   private List<String> extraFiles = new ArrayList<String>();
   private transient boolean addPlateAcquisition = false;
@@ -127,6 +129,10 @@ public class CQ1Reader extends OMETiffReader {
       for (int c=0; c<getEffectiveSizeC(); c++) {
         if (c < channelNames.size()) {
           metadataStore.setChannelName(channelNames.get(c), i, c);
+          // explicitly set brightfield channels to white
+          if (channelNames.get(c).equals(BRIGHTFIELD_NAME)) {
+            metadataStore.setChannelColor(new Color(255, 255, 255, 255), i, c);
+          }
         }
       }
     }
@@ -297,7 +303,7 @@ public class CQ1Reader extends OMETiffReader {
             if (enabled == null || enabled.equalsIgnoreCase("true")) {
               String mode = channel.getAttribute("icm:Method");
               if (mode.equalsIgnoreCase("brightfield")) {
-                names.add("Brightfield");
+                names.add(BRIGHTFIELD_NAME);
               }
               else {
                 String lightSource = getAttribute(


mabruce (approved):
With this PR, an imported CQ3000 plate now has the `Brightfield` channel rendering as white. The OME-XML also contains the expected `Color="-1"` value.

Image

sbesson (commented):
Using a sample dataset generated by a CQ3000 instrument and comparing the metadata of an OME-Zarr 0.5 dataset converted with 0.13.0-rc1 and with a build of this PR, the only changes in the Zarr attributes are in the color value under the omero metadata:

```diff
$ diff -r Plate2_Run1_120826.zarr/ Plate2_Run1_120826_pr340.zarr/
diff -r Plate2_Run1_120826.zarr/A/1/0/zarr.json Plate2_Run1_120826_pr340.zarr/A/1/0/zarr.json
56c56
<           "color" : "FF0000",
---
>           "color" : "FFFFFF",
diff -r Plate2_Run1_120826.zarr/A/1/1/zarr.json Plate2_Run1_120826_pr340.zarr/A/1/1/zarr.json
56c56
<           "color" : "FF0000",
---
>           "color" : "FFFFFF",
diff -r Plate2_Run1_120826.zarr/A/1/2/zarr.json Plate2_Run1_120826_pr340.zarr/A/1/2/zarr.json
56c56
<           "color" : "FF0000",
---
>           "color" : "FFFFFF",
```

as well as the `Channel.Color` element of the OME-XML metadat

```diff
$ diff <(/opt/omero/apps/bftools/current/xmlindent Plate2_Run1_120826.zarr/OME/METADATA.ome.xml) <(/opt/omero/apps/bftools/current/xmlindent Plate2_Run1_120826_pr340.zarr/OME/METADATA.ome.xml) 
1781c1781
<          <Channel AcquisitionMode="SpinningDiskConfocal" ContrastMethod="Brightfield" ID="Channel:0:0" IlluminationType="Transmitted" Name="Brightfield" SamplesPerPixel="1">
---
>          <Channel AcquisitionMode="SpinningDiskConfocal" Color="-1" ContrastMethod="Brightfield" ID="Channel:0:0" IlluminationType="Transmitted" Name="Brightfield" SamplesPerPixel="1">

...
<          <Channel AcquisitionMode="SpinningDiskConfocal" ContrastMethod="Brightfield" ID="Channel:382:0" IlluminationType="Transmitted" Name="Brightfield" SamplesPerPixel="1">
---
13746c13746
<          <Channel AcquisitionMode="SpinningDiskConfocal" ContrastMethod="Brightfield" ID="Channel:383:0" IlluminationType="Transmitted" Name="Brightfield" SamplesPerPixel="1">
---
>          <Channel AcquisitionMode="SpinningDiskConfocal" Color="-1" ContrastMethod="Brightfield" ID="Channel:383:0" IlluminationType="Transmitted" Name="Brightfield" SamplesPerPixel="1">
```

Assigning white as the color is consistent with the expectations of the OME schema when the attribute is missing https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Channel_Color. It highlights a discrepancy between the OME schema and the default assignment on OMERO import when the metadata is missing. This discrepancy is historical and outside the scope of this work.

To be extra defensive, would it be possible to add a check for the presence of the channel color in addition to the channel name? This would prevent a hypothetical metadata loss scenario if brightfield channel had an explicit color set in the OME-XML.

melissalinkert:
> To be extra defensive, would it be possible to add a check for the presence of the channel color in addition to the channel name? This would prevent a hypothetical metadata loss scenario if brightfield channel had an explicit color set in the OME-XML.

See ed9267d (it was easier than I initially thought). An easy way to test is to edit the OME-XML in the main .ome.tif file to include a color for a brightfield channel:

```
$ tiffcomment plate.ome.tif > plate.ome.xml
# edit plate.ome.xml add a "Color" attribute (not -1) to any "Channel" with "ContrastMethod" set to "Brightfield"
$ tiffcomment -set plate.ome.xml plate.ome.tif
$ bioformats2raw --no-tiles plate.ome.tif plate.zarr
```

The chosen color should then be present in the `OME/METADATA.ome.xml` as well as under `omero` in the Zarr metadata.

mabruce:
I added a color to one Brightfield channel in the OME-XML, injected it back into the OME-TIF, converted via the latest build, and imported into OMERO Plus. The value set in OME-XML (`-8388090`) matches the imported color (`FF8002`). The Zarr metadata matches as well:
```console
$ jq '.omero.channels[0] | {label,color}' /path/to/CQ3000_edit.zarr/A/1/0/.zattrs
{
  "label": "Brightfield",
  "color": "FF8002"
}
```

Added for clarity: the unedited brightfield channels remained white, and the fluorescence channels with existing colors were unaffected.

sbesson (approved):

